PR #317, opened 2026-04-05
feat: terminal dispatch — agent works in visible tmux via Claude Code CLI (AI-181)

jonhill90:
## Summary
Replaces the hidden tool-use loop with visible terminal work when enabled. Instead of agent → LLM → hidden subprocess, the agent now:

1. Writes the user's task to `/workspace/current-task.md` (with SOUL.md + RULES.md context)
2. Runs `claude --print < /workspace/current-task.md | tee /workspace/.claude-result` in the tmux session via `tmux send-keys`
3. All Claude Code commands, file edits, and output are **visible in xterm.js**
4. Polls for completion sentinel, captures result, sends back as chat response

### Activation
- Set `AGENT_USE_TERMINAL=1` env var on the agentbox container
- Requires `tmux` and `claude` CLI in the container (both already installed)
- **Opt-in**: agents without the env var use the legacy tool-use loop unchanged

### Key design decisions
- **Feature-flagged**: `AGENT_USE_TERMINAL` env var — zero risk to existing agents
- **Legacy preserved**: tool-use loop stays as fallback, no behavioral change unless opted in
- **Thinking callbacks**: UI gets periodic "Working in terminal..." updates during long tasks
- **Timeout**: 10 minute ceiling per task, returns partial output on timeout

Closes AI-181

## Test plan
- [x] 7 new tests in `TestTerminalDispatch` class
  - [x] Feature detection: disabled by default, enabled with env+tmux+claude, disabled without claude, disabled without tmux
  - [x] End-to-end: writes task file with soul/rules/task, calls tmux send-keys, delivers callback
  - [x] Error: returns error callback when no user message found
  - [x] Fallback: legacy tool-use path works when terminal disabled
- [x] All 180 agentbox tests pass
- [ ] Post-deploy: set `AGENT_USE_TERMINAL=1` on agent, send chat message, verify terminal shows Claude Code working

### DIFF
--- a/services/agentbox/app/chat.py
+++ b/services/agentbox/app/chat.py
@@ -1,13 +1,14 @@
-"""Chat work handler — inference call + tool-calling loop + callback delivery.
+"""Chat work handler — dispatches tasks to Claude Code CLI in tmux.
 
 Agentbox is the sole owner of system prompt assembly (§7.5).
 API sends only structured data: messages array, model, callback info.
 Agentbox reads identity from mounted files (SOUL.md + RULES.md with
 baked-in skill instructions) and prepends the system prompt.
 
-When tools are enabled, the handler runs an iterative loop: call LLM,
-execute any requested tool calls, feed results back, repeat until the
-LLM produces a final text response or the iteration limit is reached.
+Primary path (AI-181): writes the user's task to /workspace/current-task.md,
+then runs `claude` in the tmux session so all work is visible in the
+live terminal (xterm.js). Falls back to the legacy tool-use loop when
+claude CLI is not available or AGENT_USE_TERMINAL is not set.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


Snailflyer:
The opt-in terminal dispatch path would benefit from one extra post-deploy acceptance check: prove that the visible tmux session is the execution authority, not just a display surface.

Suggested smoke test after enabling `AGENT_USE_TERMINAL=1`:

- write a unique marker into `/workspace/current-task.md`
- start terminal dispatch from chat
- capture target tmux pane id and output cursor before dispatch
- verify Claude Code starts from that same pane
- verify the completion sentinel/result file came from the same terminal run
- report `terminal_run_consumed`, `rendered_but_not_started`, `wrong_pane`, or `timeout`

That keeps the key claim tight: hidden tool-loop fallback and visible tmux execution are different authority models.

Faryo uses the same acceptance boundary for phone/browser control: the action only counts when the original live tmux-backed session advances:
https://github.com/Snailflyer/faryo/releases/download/v1.0.7/faryo-public-redacted-same-session-handoff-walkthrough-20260528-0120.gif

PR #456, opened 2026-04-10
fix: use persistent event loop for browser tool calls

jonhill90:
## Summary
The browser tool (Playwright) was broken across multiple tool calls because `asyncio.run()` creates and destroys a new event loop per call, killing the browser page between calls.

**Before**: Each `asyncio.run(execute_tool_call(...))` in `_run_tool_loop` created a fresh event loop. When the browser tool launched Playwright, it bound to that loop. When the next tool call used `asyncio.run()` again, the previous loop was destroyed, killing the browser page. The `_browser_page` global pointed to a dead object.

**After**: A persistent event loop runs in a dedicated daemon thread (`tool-event-loop`). All async tool calls are dispatched to it via `asyncio.run_coroutine_threadsafe()`. The browser page survives across tool calls, and cached screenshots remain valid for the `/screenshot` endpoint.

## Changes
- `services/agentbox/app/chat.py`:
  - Added `_get_tool_loop()` — creates a persistent `asyncio.AbstractEventLoop` in a daemon thread
  - Added `_run_async(coro)` — dispatches coroutines to the persistent loop via `run_coroutine_threadsafe`
  - Replaced `asyncio.run(execute_tool_call(...))` with `_run_async(execute_tool_call(...))`

## How screenshot caching works (already in place)
The `/screenshot` endpoint serves cached bytes (`_browser_last_screenshot`), not a live `page.screenshot()` call. The cache is updated after every navigate/click/screenshot browser action via `_capture_live_screenshot()`. This is cross-loop safe since it's plain `bytes`.

## Test plan
- [x] `pytest tests/ -x -q` — 181/181 tests pass
- [ ] Rebuild agentbox, start agent, send navigate command, verify `/screenshot` returns cached screenshot
- [ ] Verify browser survives across multiple tool calls in a single conversation

### DIFF
--- a/services/agentbox/app/chat.py
+++ b/services/agentbox/app/chat.py
@@ -19,6 +19,7 @@ import logging
 import os
 import shutil
 import subprocess
+import threading
 import time
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


jonhill90:
Duplicate of #455 which already passes CI

PR #535, opened 2026-07-26
refactor(edge): move ACME DNS-01 from dns-manager to lego's Cloudflare provider

jonhill90:
## Plan

`hill90.com` is moving to Cloudflare. Seven hosts — `vault`, `portainer`, `grafana`, `storage`, `admin`, `traefik`, `litellm` — resolve into the Tailscale range `100.64.0.0/10`, so an HTTP-01 validator cannot reach them and their certificates come from DNS-01.

That DNS-01 path ran through `services/dns-manager`, a Flask shim translating lego's `httpreq` webhook calls into Hostinger DNS API writes. Left in place across the nameserver change it would write to a zone Hostinger no longer serves. Nothing breaks that day — Let's Encrypt renews at 30 days remaining on a 90-day cert, so it surfaces weeks later as expired certificates, one host at a time.

Traefik embeds [lego](https://go-acme.github.io/lego/), which has a first-class Cloudflare provider, so the shim is deleted rather than ported.

Verified against <https://go-acme.github.io/lego/dns/cloudflare/> (fetched 2026-07-26). The documentation states the token needs:

> "Zone / Zone / Read" and "Zone / DNS / Edit"

Provider code `cloudflare`, credential `CF_DNS_API_TOKEN`, defaults `CLOUDFLARE_POLLING_INTERVAL` 2s / `CLOUDFLARE_PROPAGATION_TIMEOUT` 120s / `CLOUDFLARE_TTL` 120s. None are overridden — the defaults suit this zone.

### Which file actually runs

There were **two** Traefik static configs and both had `provider: httpreq`:

| File | Status |
|---|---|
| `platform/edge/traefik.yml` | **This is the one that runs.** Mounted by `docker-compose.infra.yml:60`. |
| `infra/ansible/playbooks/09-traefik.yml` | Templates a second copy. Imported by **no** runner — `bootstrap.yml:49` explicitly skips it. |

Changing only the Ansible playbook would have left production still routing DNS-01 through `dns-manager` while looking done. Both are changed, and the playbook now carries a comment saying which is authoritative.

### Traefik is pinned to v2.11

On v2.11 the `dnsChallenge` keys are flat — `provider`, `delayBeforeCheck`, `resolvers`, `disablePropagationCheck`. The `propagation.*` block in current Traefik docs is **v3 only** and would parse without error while doing nothing. So `delayBeforeCheck: 30s` and the `1.1.1.1`/`8.8.8.8` resolvers are unchanged; only `provider` and the environment moved.

### What was kept, and why

Hostinger is still the VPS host and still the mail provider. The MX/SPF/DKIM records point at Hostinger mail permanently.

**Kept:** `scripts/hostinger.sh` (a unified VPS+DNS CLI that `scripts/vps.sh` calls for rebuilds at `:112`, `:126`, `:137`), `tests/scripts/hostinger.bats`, the Makefile `dns-*` targets, and `HOSTINGER_API_KEY` — reassigned to vault path `secret/infra/vps` and no longer referenced by compose.

**Removed:** `services/dns-manager/` (3 files), both service blocks, the playbook tasks that copied its source to the VPS, and `HTTPREQ_*`. `VERSION` also left the env surface — `image: hill90/dns-manager:${VERSION:-latest}` was its only `${VERSION}` consumer in compose.

`services/` no longer exists at all; `dns-manager` was its last occupant. `README.md`, `SPEC.md` and `PRD.md` all still described it as live infrastructure — `SPEC.md:51` specifically argued `services/` must not be deleted wholesale.

## Risks

**A blank `CF_DNS_API_TOKEN` is invisible until renewal.** lego validates the credential only at `present` time. Traefik starts, loads the resolver, logs `Testing certificate renew...`, and passes `deploy.sh verify` with an empty token.

The second commit here fixes a path into exactly that state: `vault.sh`'s `get_secret()` is an unguarded `grep | cut` returning empty for an absent key, and `bao kv put KEY=` accepts it — so `make vault-seed` run before the token was minted would print *"All secrets seeded to KV v2!"* while installing a blank credential, and `sync-to-sops` would write the blank back into `prod.enc.env` as a real key. Seeding now validates every required key in the parent shell before any write. (The first attempt at this fix did not work — `die` inside `$(get_secret ...)` exits only the subshell. Caught by running it.)

**`check_secrets_schema.py` cannot see vault-path drift.** It cross-references schema ↔ compose ↔ `prod.enc.env.example` and never opens `vault.sh` or `_common.sh`. Its "all checks passed" says nothing about whether a declared `vault_path` is actually seeded. `secret/infra/vps` was declared-but-unseeded until this PR.

**`ACME_CA_SERVER` and `letsencrypt_env` are inert — pre-existing, NOT fixed here.** Compose passes the CA as `--certificatesresolvers.*.acme.caserver=` CLI flags, but Traefik v2 discards CLI flags when a static config file is mounted, and `platform/edge/traefik.yml` defines no `caServer`. Both resolvers therefore use **production** Let's Encrypt, and nobody has ever been able to test issuance against the staging CA on the live path. `certificates.md` claimed the opposite; corrected here and flagged as a trap. Filed separately — deliberately not fixed in this PR.

**Blast radius of the token.** Zone/DNS/Edit permits more than TXT writes; Cloudflare offers no TXT-only scope. Scope to the `hill90.com` zone only and never use a Global API Key.

## Rollback

`git revert` both commits and redeploy the edge stack. This restores `services/dns-manager` and the `httpreq` provider, which remains correct for as long as the nameservers still point at `ns1/ns2.dns-parking.com`. There is no state migration and no volume change; `traefik-certs` is untouched and already-issued certificates are unaffected.

After the nameservers move, revert is no longer a valid rollback — the shim would write to a zone Hostinger no longer serves. Roll forward instead.

## Validation Evidence

```
$ python3 scripts/checks/check_env_surface.py
Configuration surface consistent: 32 variable(s) used, all documented and all defaulted to production values.

$ python3 scripts/checks/check_secrets_schema.py
Secrets schema validation: all checks passed

$ python3 scripts/checks/check_md_links.py
Markdown link check passed: no missing local links found.

$ python3 scripts/checks/check_volume_names.py
All 3 compose files have explicit volume names.

$ bats tests/scripts/
249 ok, 0 failures

$ docker compose -f deploy/compose/prod/docker-compose.infra.yml config
exit 0
```

The three new `vault.bats` tests were confirmed **red against 7aefece and green against b5e85b8**, not merely green after the fix.

An adversarial audit subagent reviewed the first commit and found the blank-credential path, the unseeded vault path, and the `caServer` claim. Its report also independently re-verified the Traefik v2.11 key semantics and the lego provider details against the official docs.

## ⚠️ Unproven — pending runtime proof

**No certificate has been issued against any ACME server on this path.** A green CI run does not prove this change; the thing that can fail is issuance, and it fails quietly.

`CF_DNS_API_TOKEN` is wired through every surface *except* `infra/secrets/prod.enc.env` itself — the token has not been minted yet.

### Pending proof method

Do **not** force issuance on a real service hostname against production Let's Encrypt. Those hosts have working certificates, and a failed attempt risks the duplicate-certificate limit on a host that currently renews fine.

Instead, once the token exists:

1. Create a throwaway hostname with no certificate history — `acmetest.hill90.com` — with a temporary A record and a temporary Traefik router using `certresolver=letsencrypt-dns`. A fresh name has no duplicate-cert history, so the only rate-limit budget at risk is on a name nothing depends on.
2. Watch the `_acme-challenge` TXT record **appear and then be removed** in the Cloudflare zone. Both halves matter: a provider that creates but never cleans up leaves the zone accumulating junk.
3. Read the Traefik log for the actual challenge result. Do not infer success from the container being healthy — Traefik stays healthy through a failed renewal.
4. Tear down the temporary A record and router.

### Required from the repo owner

Cloudflare dashboard → My Profile → API Tokens → Create Token → Custom token:

- **Permissions:** `Zone / Zone / Read` **and** `Zone / DNS / Edit`
- **Zone Resources:** Include → Specific zone → `hill90.com`
- **Not** a Global API Key

## Sequencing

The nameserver change at the Hostinger registrar is the repo owner's action and happens **after** this lands. The registrar has not been touched.

One item must land before that cutover and is **not** in this PR: `hostinger.sh dns_sync` still writes A records to the Hostinger zone and is invoked by `config-vps.yml:165-170` on every VPS rebuild. After the nameserver move that writes into a dead zone and reports success — the same silent-failure shape, on the record path instead of the certificate path. Tracked as a separate PR by explicit scope decision.


jonhill90:
Filed the `caServer` defect separately as **[DW-72](https://linear.app/debatewho/issue/DW-72/acme-ca-server-and-letsencrypt-env-are-inert-staging-ca-has-never-been)** — `ACME_CA_SERVER` and `letsencrypt_env` are inert because Traefik v2 discards CLI flags when a static config file is mounted, so the staging CA has never been testable on the live path. Not fixed here by explicit scope decision.

jonhill90:
Correction to my earlier comment: I filed the `caServer` defect in Linear, which #533 retired as the tracker earlier today. Re-filed as **#537** in this repository. The Linear copy (DW-72) is cancelled and points here.

Also rebased onto `main` (now includes #533). Re-verified after the rebase: 249/249 bats, all four checks pass.

jonhill90:
Follow-up **#540** opened: ports `dns_sync` from the Hostinger API to Cloudflare. It depends on this PR for the `CF_DNS_API_TOKEN` plumbing, so **#535 merges first**. Both must be in before the registrar nameserver change.

jonhill90:
**Token landed on this branch.** `CF_DNS_API_TOKEN` is now in `infra/secrets/prod.enc.env` on `refactor/cloudflare-dns`, so it arrives with the PR that consumes it. Putting it in #540 instead would have left a window where this PR deploys with the token unset — the compose default is blank — and lego fails every issuance silently.

Verified: file decrypts, 65 keys intact, token reports `active`, and listing zones returns only `hill90.com` so the scoping holds.

The Cloudflare **API** path is now proven end-to-end in #540 against the live zone (create, update-preserving-other-fields, idempotency, whole-zone diff showing nothing else touched). That does not prove *this* PR — issuance is the thing that can fail here, and it still needs Traefik recreated on the VPS.

jonhill90:
## Prepared, not run — and one finding that changes the order of operations

The token exists and the Cloudflare API path is proven (#540, live zone). What remains for *this* PR is certificate issuance, which needs Traefik recreated on the VPS. That is a deploy, so it is gated on Jon and I have stopped at the point where it would run.

While preparing it I hit something that has to be settled first.

### ⚠️ The acmetest proof cannot succeed today, and deploying this PR before the nameserver move would break issuance

Let's Encrypt validates DNS-01 by querying the **authoritative** nameservers. They are still Hostinger's:

```
$ dig NS hill90.com +short
ns2.dns-parking.com.
ns1.dns-parking.com.

$ cloudflare zone status
{"status":"pending","name_servers":["adi.ns.cloudflare.com","art.ns.cloudflare.com"]}
```

So if this PR is deployed while the zone is still `pending`, lego writes the `_acme-challenge` TXT into a Cloudflare zone **nobody is serving**, Let's Encrypt queries Hostinger, never sees it, and every issuance fails. The current `dns-manager` path works today precisely *because* Hostinger is still authoritative.

The brief had this the right way round for the **code** — both PRs land before the registrar change — but it did not separate landing from deploying. They must go in opposite orders:

| Step | Who | Why |
|---|---|---|
| 1. Merge #535, then #540 | — | Code only. Nothing deployed, nothing breaks. |
| 2. Move nameservers at the Hostinger registrar to `adi.ns.cloudflare.com` / `art.ns.cloudflare.com` | **Jon** | Makes the Cloudflare zone authoritative. |
| 3. Wait for `dig NS hill90.com` to return Cloudflare and zone status to read `active` | — | Until this, DNS-01 cannot validate through Cloudflare. |
| 4. Deploy the edge stack (recreate Traefik) | **Jon-gated** | Traefik picks up `provider: cloudflare` and the token. |
| 5. Run the acmetest proof below | — | First real issuance on the new path. |

**Deploying at step 4 before step 2 is the failure mode.** It is the mirror image of the one this migration exists to fix.

### The window between steps 2 and 4 is safe, and here is the budget

Between the nameserver move and the deploy, the old `dns-manager` is still running and would write to a zone Hostinger no longer serves. That is harmless *provided no renewal fires in the gap*, because Let's Encrypt only renews at 30 days remaining:

```
traefik/portainer/grafana   expires 2026-09-12   renewal starts 2026-08-13   ->  18 days of runway
vault                       expires 2026-10-24   renewal starts 2026-09-24   ->  60 days of runway
```

**18 days.** The binding constraint is `traefik`, `portainer` and `grafana`. Steps 2–4 should complete well inside that, and there is no need to rush them into the same maintenance window.

(These four are the only DNS-01 hosts currently serving a real certificate. `admin`, `litellm` and `storage` have `_acme-challenge` TXTs in the zone but serve Traefik's default cert — see #538, which predates this work.)

---

## The acmetest procedure (to run at step 5)

Throwaway hostname with no certificate history, so the only rate-limit budget at risk is on a name nothing depends on. Not a real service hostname — those have working certs and a failed attempt risks the duplicate-certificate limit.

**Prerequisites:** `dig NS hill90.com +short` returns the Cloudflare nameservers, and the zone reads `active`.

### 1. Temporary A record

```bash
bash scripts/cloudflare.sh dns get                    # confirm the token works from the repo
# create acmetest.hill90.com -> the Tailscale IP, via the dashboard or a direct API call
```

Deliberately **not** added to `MANAGED_RECORDS` — it is temporary, and the allowlist is the safety property.

### 2. Temporary router

Add to the edge stack a throwaway router on `acmetest.hill90.com` using `certresolver=letsencrypt-dns`, pointing at any existing service. Recreate Traefik.

> **Staging is not available.** `ACME_CA_SERVER` is inert (#537) — Traefik v2 discards CLI flags when a static config is mounted — so this necessarily runs against **production** Let's Encrypt. That is why a fresh hostname matters: 5 failed authorizations per hostname per hour, and `acmetest` has no history to burn.

### 3. Watch both halves of the TXT lifecycle

```bash
watch -n5 'dig TXT _acme-challenge.acmetest.hill90.com +short @adi.ns.cloudflare.com'
```

The record must **appear and then be removed**. A provider that creates but never cleans up leaves the zone accumulating junk — and the zone already carries seven stale `_acme-challenge` TXTs from the dns-manager era, which is what that failure looks like after a while.

### 4. Read the Traefik log for the actual result

```bash
ssh deploy@remote.hill90.com 'docker logs traefik --tail 200 | grep -iE "acme|certificate|challenge|cloudflare"'
```

**Do not infer success from the container being healthy** — Traefik stays healthy through a failed renewal. That is the whole reason this proof exists.

### 5. Confirm the certificate

```bash
echo | openssl s_client -connect acmetest.hill90.com:443 -servername acmetest.hill90.com 2>/dev/null \
  | openssl x509 -noout -issuer -dates
```

### 6. Tear down

Remove the temporary router, recreate Traefik, delete the `acmetest` A record and any leftover `_acme-challenge.acmetest` TXT. Confirm the zone is back to 32 records.

### 7. Then, separately

Once issuance is proven, the seven stale `_acme-challenge` TXTs carried over from Hostinger can be removed — the brief asked for that to be confirmed from the ACME flow rather than assumed, and step 3 is that confirmation.

---

**Nothing above has been run.** No deploy has been performed and no VPS-mutating command has been issued.

PR #545, opened 2026-07-27
fix(traefik): reject silently-ignored configuration instead of accepting it

jonhill90:
> ## ⛔ Merge order
>
> Branched from `refactor/cloudflare-dns` (#535) because it edits the same Traefik files. **Merge #535 first.** Independent of #540.
>
> Nothing here needs the nameserver cutover. It fixes defects that exist on `main` today.

## Plan

Three issues — #537, #541, #543 — with one shape: **configuration that is silently ignored rather than rejected.** Fixed as one PR against that theme.

### #537 / #543 — `ACME_CA_SERVER` was inert, then dangerous either way

The Traefik v2.11 docs state the mechanism plainly:

> "There are three different, mutually exclusive (i.e. you can use only one at the same time), ways to define static configuration options in Traefik"

A config file is mounted, so the compose `command:` flags —
`--certificatesresolvers.*.acme.caserver=${ACME_CA_SERVER}` — were interpolated by Compose and then discarded by Traefik. `ACME_CA_SERVER` and the Ansible `letsencrypt_env` did nothing. Both resolvers silently used Traefik's default (production Let's Encrypt), and staging was untestable.

`platform/edge/traefik.yml` becomes `traefik.yml.tmpl`, rendered by `scripts/render-traefik-config.sh` into a gitignored `traefik.generated.yml` that compose mounts. That makes the variable authoritative.

It is **required, with no default**, because both defaults are dangerous in opposite directions — a staging default silently replaces every certificate with an untrusted one on any deploy without secrets; a production default burns real rate limits from an unconfigured environment. A deploy that stops is better than either.

### #541 — deleted, not renamed

`09-traefik.yml` referenced `letsencrypt` while defining `letsencrypt-http` and `letsencrypt-dns`, pointed the HTTP-01 store at a file the volume does not contain, and declared its own `traefik` container next to the compose one. `10-portainer.yml` did the same for portainer.

Checked whether the playbook should exist at all: **nothing imports either.** `bootstrap.yml` imports 01-08, 11, 12; the only `ansible-playbook` invocation is `vps.sh:211` → `bootstrap.yml`; no workflow, Makefile target or runbook names them. They are not the bootstrap path — they are dead code that would break TLS if run.

Deleted, because two divergent declarations of one container is what produced the bug and keeping both in step forever is not a fix. `bootstrap.yml` and `docker-compose.infra.yml` now each state which is the **bootstrap path** and which is the **running path**.

## Risks

**⚠️ Recovering from an accidental staging issuance is expensive**, which is why the old staging default mattered more than it looks. Traefik will not reissue a certificate it considers valid, and a staging certificate *is* valid — correctly signed, unexpired, merely signed by a CA no browser trusts. Redeploying with the right CA does not fix it.

Recovery means clearing the stores so Traefik requests new ones. They live in the `traefik-certs` Docker volume, **root-owned**, and `acme-dns.json` holds **all four DNS-01 certificates in one file** — traefik, portainer, grafana, vault. There is no way to clear one host without clearing the others; every one is reissued. `acme.json` separately holds the HTTP-01 certificate for `auth`, and the stored ACME account registrations are production registrations, a further mismatch when the resolver is pointed at staging.

So one unconfigured deploy costs a full reissue of every certificate on the host, by hand, against root-owned files. That cost is documented in `certificates.md`.

**Deploying still requires the secrets store to hold the right CA.** `ACME_REQUIRE_PRODUCTION=1` turns that into a hard failure rather than a silent staging issuance, but it cannot fix a wrong value — it refuses to proceed and tells you where to correct it.

## What the audit caught, and what it changes

An adversarial audit of the first commit found that **the fix would have taken the stack down**, and that three of its six guards did not constrain what they claimed. Both are fixed in `58fc88b1`.

**`sops exec-env` swallows failures.** It runs its command in a new shell that does not inherit `deploy.sh`'s `set -e`, and returns 0 regardless. Proven:

```
$ sops exec-env ... 'echo step1; bash -c "exit 7"; echo STILL RAN'
step1
STILL RAN
exec-env exit=0
```

So a failed render was swallowed and `docker compose up` ran anyway. Docker materialises a missing bind-mount source as a **directory**, Traefik receives a directory at `/etc/traefik/traefik.yml` and will not start, and every routed service goes down while the deploy reports success — and every retry fails at the same point for a new reason and swallows that too. After the fix the same test exits 7.

**The CA choice was still not the caller's.** Both secret mechanisms overwrite a caller-set value — `sops exec-env` replaces it, and `_common.sh:183` does an unconditional `set -a; source`:

```
$ ACME_CA_SERVER=https://CALLER-WINS/directory sops exec-env ... 'echo $ACME_CA_SERVER'
https://acme-v02.api.letsencrypt.org/directory
```

So `Makefile:136` and `.github/workflows/deploy-infra.yml:80` both exported the production URL and neither chose anything — including the production CI path. `make deploy-infra` and `make deploy-infra-production` were behaviourally identical, and the "Are you sure?" prompt guarded nothing.

That mattered *more* after my change, not less: making the variable authoritative without fixing this means a stale staging value in vault would now be obeyed, and CI could not stop it. Both exports are removed because they lied, replaced by `ACME_REQUIRE_PRODUCTION=1` — not a secret, so the store cannot override it.

## Validation Evidence

Every guard was proven to fire against a real defect, not merely to pass.

```
$ bats tests/scripts/                                    274 ok, 0 failures
$ shellcheck -S error render/preflight/deploy/validate   clean
$ check_env_surface / check_secrets_schema / check_md_links / check_volume_names   pass
$ docker compose -f deploy/compose/prod/docker-compose.infra.yml config   exit 0
$ bash scripts/validate.sh traefik                       all ✓, caServer 2/2
```

**Render, run against real inputs:**

| Input | Result |
|---|---|
| unset / empty / `not-a-url` | rejected, exit 1, no file created |
| staging | renders, `STAGING` + `UNTRUSTED` warnings |
| production | renders, 2 caServer lines, no placeholders, valid YAML |
| unknown CA | renders, warns it is not Let's Encrypt |
| `ACME_REQUIRE_PRODUCTION=1` + staging | **refused**, no file created |
| failing render over an existing good file | original left byte-intact |
| directory at the output path | cleared, renders |

**Preflight:** rejects missing, empty, and directory; accepts a correct render. Proven to catch a directory through `validate.sh` too.

**Guards proven red before green:**
- Recreated `09-traefik.yml` → orphan test goes red (the previous version passed, because this commit's own comment names the file).
- Changed a resolver ref to defaultless `${VAR}` → red. Pointed one at a nonexistent resolver → red.
- Removed a `caServer` → `validate.sh` reports 1 of 2. Added a stray placeholder → caught.

**The old guards encoded the bug.** `validate.bats` and `validate.sh` both asserted `traefik.yml` contained no `${VAR}` — right about the danger, but unable to distinguish template from mounted file, which is exactly why the CA had to stay hardcoded. Both now check the rendered output.

**A guard I added found a defect I had just introduced.** `check_secrets_schema.py` counted variables named in *comments* as compose references; fixing that immediately surfaced a stale `compose_ref` for `ACME_CA_SERVER` left by this very change.

## Rollback

`git revert`. `rollback.sh service_paths` returns `platform/edge/`, so `git checkout <ref> -- platform/edge/` restores `traefik.yml` and the rolled-back compose mounts it. No state migration; `traefik-certs` is untouched and issued certificates are unaffected.

One manual step if reverting on the VPS: remove `platform/edge/traefik.generated.yml`, which is untracked and survives `git reset --hard`.

## Not proven

No deploy has been run. These changes affect the deploy path, and a deploy needs Jon. The render, preflight and validate paths are proven by direct execution; the end-to-end deploy is not.


jonhill90:
## Fourth member of the family: the `tailscale-only` allowlist

Added in `9869e044`. `platform/edge/dynamic/middlewares.yml` allowed two ranges — the Tailscale CGNAT range and the Docker bridge gateway `172.18.0.1/32`. The gateway entry's own comment said it was there because proxied requests could arrive with the gateway as their source after a container-runtime restart.

**I checked that claim rather than assuming it.** The access log window covers a runtime restart (2026-07-21 03:02 UTC, confirmed via `systemctl show docker`), and every Tailscale request after it arrived with its real CGNAT source address — never the gateway. The IPv4 DNAT rules that preserve the source address are present on the host, which is the mechanism that makes that so.

**The stronger argument is that it would not be a workable control even if the symptom returned.** The gateway address carries no information about where a request originated: anything the container runtime rewrites arrives with that same source, so a rewritten request cannot be attributed to the private network. An allowlist entry that cannot make that distinction is not narrowing anything — which is exactly this PR's theme, a rule that reads as a working control while being silently permissive.

**It fails closed.** If the restart symptom recurs, admin routes return 403 until the runtime's forwarding rules settle — a temporary lockout, recoverable over SSH, rather than a permanently widened allowlist. The file says not to re-add a bare gateway CIDR to fix it, and to match on something an external client cannot present.

Guards, proven to fire by re-adding the entry:

```
not ok 26 tailscale-only allows only the Tailscale CGNAT range
not ok 27 tailscale-only carries no bridge, private or loopback CIDR
Checking tailscale-only allowlist scope... ✗ Allowlist contains a private or bridge CIDR
```

A third test records that portainer, grafana and vault use this middleware **without** basic auth, so the allowlist is their only network-level scoping and must stay narrow.

Related: #542 tracks a DNS decision that would change what reaches this path, and is worth settling before it does.

277/277 bats, all four repo checks pass, shellcheck clean.

jonhill90:
## Fifth member of the family: #539, SSH hardening

Added in `9e72277a`. Same shape as the rest of this PR — a step that runs, reports success, and has no effect.

### Mechanism, confirmed on the host

The role wrote `PasswordAuthentication no` with `lineinfile` into `/etc/ssh/sshd_config`. sshd takes the **first** value it sees for a keyword, and the stock AlmaLinux config carries `Include /etc/ssh/sshd_config.d/*.conf` at **line 15** while the role's value landed at **line 64**. Cloud images ship `50-cloud-init.conf` containing `PasswordAuthentication yes`, so the include resolved it first and the hardened value was never reached.

```
sshd_config:15   Include /etc/ssh/sshd_config.d/*.conf
sshd_config.d/50-cloud-init.conf:1   PasswordAuthentication yes   ← wins
sshd_config:64   PasswordAuthentication no                        ← never reached
```

`PermitRootLogin no` and `MaxAuthTries 3` **did** apply, because nothing overrides those. Two thirds of the role worked, which is why this survived a full rebuild. And cloud-init regenerates its drop-in, so it would come back after any future one.

### The fix

Hardening moves to `/etc/ssh/sshd_config.d/00-hill90-hardening.conf` — named `00-` so it is read before `50-cloud-init.conf` and wins. cloud-init is also told not to manage the setting, so the two do not disagree on disk.

`sshd -t` validates the **full effective** config before any restart, and an invalid drop-in is **removed** rather than left in place — otherwise the next restart by anything would fail, and the only route in is the tailnet.

Then it is verified, which is the part whose absence let this persist. The role reads `sshd -T` and asserts the effective values match intent. `scripts/verify-ssh-hardening.sh` does the same read-only, without ansible, so the claim is checkable without a deploy. Run against the live host it correctly reports the current state:

```
passwordauthentication            yes  ← expected no
permitrootlogin                   no
pubkeyauthentication              yes
kbdinteractiveauthentication      no
permitemptypasswords              no
maxauthtries                      3
✗ 1 setting(s) do not match the intended hardening.   exit=1
```

### fail2ban

Installed, configured and started with `ignore_errors: yes` on all three tasks. It is in **none** of the repositories enabled on the host (BaseOS, AppStream, CRB, Extras, Docker CE, Tailscale), so every step failed, every failure was ignored, and the play reported success while the binary was absent and the docs claimed the service was enabled.

Tasks removed rather than left pretending; unused handler removed; `security.md` corrected. Restoring it needs EPEL — a deliberate decision about a third-party repo on a hardened host, and not one to make silently. Its value is also limited here: SSH is not publicly reachable and, after this change, accepts keys only.

### ⚠️ Scope and urgency

**Nothing was deployed and sshd on the host was not touched** — that is VPS-mutating and needs Jon. This lands as a playbook change; applying it is `make config-vps`.

**Exposure is bounded and this is not an emergency.** Public SSH is closed: firewalld carries no `ssh` service, and port 22 on the public address refuses connections — I verified that externally rather than taking it on trust. Access is over the tailnet only. This is a hardening gap to close deliberately, not something to rush a risky change for.

### Guards

12 new bats tests, each proven to fire: renaming the drop-in to `60-` (sorting after cloud-init's) goes red, and re-adding an `ignore_errors` fail2ban task goes red on two.

One adjacent observation, **not** changed here: the role's firewalld rich rule for `100.64.0.0/10` is not present in any active zone either, and `tailscale0` is in no firewalld zone — Tailscale filters via its own chains, which is why tailnet SSH works regardless. Touching firewall rules on a box whose only access path is the tailnet deserves its own change with Jon present.

289/289 bats, ansible syntax-check passes, all four repo checks, shellcheck clean.

PR #556, opened 2026-07-27
feat(storage): restore MinIO as a platform service with Keycloak OIDC (#530)

jonhill90:
Restores MinIO as a platform service and wires its OIDC to Keycloak. Closes the MinIO half of #530 — **but not as SSO**. Read the caveat before merging.

## ⚠️ Merge ordering — read this first

`scripts/keycloak.sh apply` resolves **every** client secret up front and aborts if any is missing. That is deliberate (an audit found it would otherwise write an empty secret over a live client and exit 0), but it makes the step **all-or-nothing**.

**Merging this without `MINIO_OIDC_CLIENT_SECRET` in SOPS would abort `keycloak.sh apply` and leave every client unconfigured — Grafana and Portainer included, undoing tonight's SSO work.** That is exactly what a missing `VAULT_OIDC_CLIENT_SECRET` cost us earlier this evening (#546).

**This is already handled**: the secret is added in this PR's first commit, verified by md5 with the trailing newline stripped, with the other five confirmed intact (70 keys, up from 69). Nothing further is required at merge time — this note exists so the ordering is understood rather than rediscovered.

## What this is

MinIO returns as a **platform primitive** — the Storage Account equivalent alongside Keycloak for Entra ID and OpenBao for Key Vault — following the shape #531 used for Keycloak and Postgres: same compose layout, same secrets plumbing, same local override story.

It also ends a live inconsistency: `storage.hill90.com` has resolved to the VPS throughout, so the platform has been advertising an object store that nothing served.

### Relevance to the app-arch lane

**hill90-app has its own `docker-compose.minio.yml`, which this supersedes.** That lane is currently deciding whether the app keeps its own object store or consumes Hill90's; this PR is what makes "consume Hill90's" the available answer. The S3 API stays on the internal network, reachable as `http://minio:9000`, which is how a tenant would consume it.

## OIDC is wired. SSO is not, and cannot be

**MinIO removed the management console from the AGPL community build in May 2025.** Verified by running releases side by side against a real Keycloak with identical configuration:

| Release | Console `loginStrategy` |
|---|---|
| `RELEASE.2025-04-22` | **`redirect`** — working Keycloak button |
| `RELEASE.2025-05-24` → `2025-09-07` | `form`, `redirectRules: null` |

So the console is **root-credential form login only**, no matter how OIDC is configured. Pinning the April release to keep the button was rejected: freezing an internet-facing service fifteen months behind on security updates is a bad trade.

What OIDC *does* provide is the **S3/STS path**, verified end to end: a Keycloak user with realm role `admin` exchanged a token via `AssumeRoleWithWebIdentity` for temporary credentials and wrote a 29-byte object. A `viewer` user could read but not write. **Do not describe this as SSO.**

Env var names came from `mc admin config set identity_openid --env` on the pinned image, not from documentation — the community docs URL now redirects to AIStor's.

## There is no consumer today

Stated because it is the most likely thing to be misread later: Loki uses `filesystem` storage, Tempo a `local` backend, and `backup.sh` has no S3 target. The primitive exists so consumers can arrive — the same reasoning that restored Keycloak and Postgres. The `MINIO_*` and `VAULT_MINIO_*` keys already in SOPS were orphans of the extracted application and are live again rather than newly minted.

## Certificate path (#538)

`storage.hill90.com` has never issued a certificate — not because ACME was broken, but because no router existed, so nothing ever asked. This adds the router and therefore triggers the **first issuance**.

The resolver is **not hardcoded**; it inherits `ADMIN_CERT_RESOLVER`, so the now-live Cloudflare DNS-01 path applies. The runbook has a watched-issuance procedure with rollback, because a failed first issuance means Traefik serves its default certificate.

## Audit findings, all fixed

A subagent audit found four production defects:

1. **`deploy.sh teardown minio` was broken** — pasted from the wrong block, never set `project_name`, and clobbered `$stack`. It skipped the backup, removed nothing, and exited 0 announcing success.
2. **`mc ready` has no timeout** — still retrying at 90s. Inside a deploy that blocks until CI's six-hour default *while holding the `deploy-prod` concurrency group*.
3. **`mc ready` is unauthenticated** — returns "ready" for bogus credentials, so the credential gate could not detect a wrong root password. Both now use `timeout 30 … mc admin info`.
4. **Empty root credentials silently started MinIO on `minioadmin:minioadmin`** with only a WARN. Now `${VAR:?…}`, which fails the deploy.

Plus four tests that passed against the exact violation they guard (a 2024 pin slipping past a 2025-only regex; a router named something other than `minio-api`; a commented-out call; alternation in a docs assertion). All re-tested by re-running the auditor's mutations.

## Verification

Cold from an empty stack: 15 containers, all health checks green, 10/10 Prometheus targets including `minio`. **325 bats / 29 pytest / shellcheck / five repo checks** pass.

## Risks

- Merging deploys MinIO to production. `deploy.sh minio` runs `minio.sh apply` (non-fatal) to provision the `admin`/`editor`/`viewer` policies; without them federated logins are rejected with "no policy found". Root access is unaffected either way.
- `prod_minio-data` was never deleted in #495, so production may find existing objects. Root credentials are reused rather than rotated for that reason — verified safe by starting a volume initialised with one credential under another.
- First certificate issuance for `storage.hill90.com`; watch it rather than assume it.

## Not done

- **MinIO console SSO** — impossible on a current release, as above.
- **Production verification** — Jon should confirm the console loads and the certificate is trusted after deploy.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -149,6 +149,12 @@ MINIO_ROOT_PASSWORD=localdev-minio-password
 # the AGPL build (removed May 2025), so this does not give you a login button.
 MINIO_OIDC_CLIENT_SECRET=local-minio-secret-0123456789
 
+# Routing. Production resolves to storage.hill90.com with no environment set.
+MINIO_HOST=storage
+MINIO_PUBLIC_URL=http://storage.localtest.me:8080
+# The token claim MinIO reads the policy name from. keycloak.sh emits it.
+MINIO_OIDC_CLAIM_NAME=minio_policy
+
 # The local Portainer admin. portainer.sh uses it to call the Portainer API, and
 # it is your way in when Keycloak is unavailable. Portainer 2.39 requires the
 # admin to be created within a few minutes of first start, using a setup token
--- a/.env.local.example
+++ b/.env.local.example
@@ -140,6 +140,15 @@ GRAFANA_OIDC_ENABLED=true
 GRAFANA_OIDC_CLIENT_SECRET=local-grafana-secret-0123456789
 PORTAINER_OIDC_CLIENT_SECRET=local-portainer-secret-0123456789
 
+# --- Object store (MinIO) ----------------------------------------------------
+# Root credentials for the local object store. Disposable; production keeps its
+# own in SOPS.
+MINIO_ROOT_USER=minioadmin
+MINIO_ROOT_PASSWORD=localdev-minio-password
+# Gates the S3/STS path against Keycloak. MinIO's console has NO SSO login in
+# the AGPL build (removed May 2025), so this does not give you a login button.
+MINIO_OIDC_CLIENT_SECRET=local-minio-secret-0123456789
+
 # The local Portainer admin. portainer.sh uses it to call the Portainer API, and
 # it is your way in when Keycloak is unavailable. Portainer 2.39 requires the
 # admin to be created within a few minutes of first start, using a setup token
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,15 @@ jobs:
         run: bats tests/scripts/
 
       - name: Validate compose files
+        # Credentials that must never fall back to a default use ${VAR:?...},
+        # so `config` refuses to render without them. That guard is deliberate:
+        # an empty MinIO root password silently starts the server on
+        # minioadmin/minioadmin with only a WARN. Supply throwaway values here —
+        # this step validates STRUCTURE, not secrets.
+        env:
+          MINIO_ROOT_USER: placeholder
+          MINIO_ROOT_PASSWORD: placeholder
+          MINIO_OIDC_CLIENT_SECRET: placeholder
         run: |
           for f in deploy/compose/prod/docker-compose*.yml; do
             echo "Validating $f..."
--- a/.github/workflows/deploy-minio.yml
+++ b/.github/workflows/deploy-minio.yml
@@ -1,0 +1,18 @@
+name: Manual Deploy Object Store (Prod)
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+concurrency:
+  group: deploy-prod
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    uses: ./.github/workflows/reusable-deploy-service.yml
+    with:
+      service: minio
+    secrets: inherit
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,6 +8,7 @@ on:
       - 'platform/data/postgres/**'
       - 'platform/observability/**'
       - 'deploy/compose/prod/docker-compose.db.yml'
+      - 'deploy/compose/prod/docker-compose.minio.yml'
       - 'deploy/compose/prod/docker-compose.auth.yml'
       - 'deploy/compose/prod/docker-compose.vault.yml'
       - 'deploy/compose/prod/docker-compose.observability.yml'
@@ -16,7 +17,7 @@ on:
       service:
         description: 'Service to deploy (or "all")'
         type: choice
-        options: [all, db, auth, vault, observability]
+        options: [all, db, auth, minio, vault, observability]
         default: 'all'
 
 permissions:
@@ -33,6 +34,7 @@ jobs:
     outputs:
       db: ${{ steps.filter.outputs.db }}
       auth: ${{ steps.filter.outputs.auth }}
+      minio: ${{ steps.filter.outputs.minio }}
       vault: ${{ steps.filter.outputs.vault }}
       observability: ${{ steps.filter.outputs.observability }}
     steps:
@@ -47,6 +49,8 @@ jobs:
             auth:
               - 'deploy/compose/prod/docker-compose.auth.yml'
               - 'platform/auth/keycloak/**'
+            minio:
+              - 'deploy/compose/prod/docker-compose.minio.yml'
             vault:
               - 'deploy/compose/prod/docker-compose.vault.yml'
             observability:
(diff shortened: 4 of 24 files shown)


jonhill90:
Reviewed and verified — **holding the merge until morning, deliberately.** Not a defect in the PR.

## What I checked
- All six key secrets present on the branch, 70 keys total: `MINIO_OIDC_CLIENT_SECRET`, `VAULT_OIDC_CLIENT_SECRET`, `GRAFANA_OIDC_CLIENT_SECRET`, `PORTAINER_OIDC_CLIENT_SECRET`, `PORTAINER_ADMIN_PASSWORD`, `CF_DNS_API_TOKEN`. So the abort-on-missing-secret trap this PR warns about is closed.
- CI green, `CLEAN`.

## Why it is held
It touches `platform/observability/prometheus/prometheus.yml`, which **is** in `deploy.yml`'s path filter. Merging therefore fires an unattended production deploy of the observability stack at ~04:00 with Jon asleep and two lanes mid-work.

Nothing needs MinIO in production tonight. The only thing that needed an answer from this PR was hill90-app's platform-dedup decision — whether the app keeps its own `docker-compose.minio.yml` or consumes Hill90's — and that is answered by this PR *existing and being green*, not by it being deployed. I am passing that answer to the `app-arch` lane directly.

The tradeoff, stated plainly: holding costs nothing tonight; merging risks debugging an observability restart at 4am against a stack whose Grafana SSO was verified working two hours ago. If the restart misbehaved, I would be fixing it while two other lanes changed the ground underneath.

## To merge in the morning
CI is green and the secrets are in place, so it should be a straight merge — expect the observability deploy to fire, and verify Grafana SSO still answers afterwards (`GF_AUTH_GENERIC_OAUTH_ENABLED` in the container, and `generic_oauth` on the login page — note `/api/frontend/settings` returns nothing useful to an unauthenticated caller and will mislead you).

jonhill90:
## Conflict resolved against current main — re-verified, still held for review

Went DIRTY when #557 merged underneath. Both branches append a clause to the same Docker provider `constraints` line in `platform/edge/traefik.local.yml` — #557 added `hill90-local` (the app stack), this branch added `hill90-local-storage` (the restored MinIO stack). That was the single conflict.

Resolved onto **main's** line rather than reapplying this branch's, keeping both clauses. Seven now — six Hill90 local stacks plus the app:

```
hill90-local-edge  hill90-local-observability  hill90-local-identity
hill90-local-platform  hill90-local-db  hill90-local-storage  hill90-local
```

The comment above the line was a pre-merge conflict *warning*; it is now a durable instruction to count the clauses when resolving, since the collision is a matter of record rather than a prediction. Dropping one silently 404s that stack's hostnames with nothing logged anywhere — a resolution can look clean and still be wrong.

### Re-verified after the merge

**Secrets — intact.** The decrypted key set is byte-identical to the pre-merge baseline, 70 keys, none added and none lost:

```
$ diff <(pre-merge keys) <(post-merge keys)   ->  identical
```

All MinIO-related secrets present, including `MINIO_OIDC_CLIENT_SECRET`:

```
MINIO_ROOT_USER  MINIO_ROOT_PASSWORD  VAULT_MINIO_ROLE_ID
VAULT_MINIO_SECRET_ID  MINIO_OIDC_CLIENT_SECRET
```

One correction worth stating: that is **five** MinIO-related keys in the encrypted file, not six. `prod.enc.env.example` has six MinIO-matching *lines*, but one is a comment explaining that `MINIO_ROOT_USER`/`MINIO_ROOT_PASSWORD` predate this PR — they outlived the service when MinIO was removed in #495. The only key this PR adds versus main is `MINIO_OIDC_CLIENT_SECRET`. Both secret CI gates pass, which is the stronger assurance than any hand count:

```
check_env_surface.py    -> 38 variable(s) used, all documented and all defaulted
check_secrets_schema.py -> all checks passed
```

**Prometheus — intact, and the deploy trigger is real.** The `minio` scrape job survives the merge unchanged:

```
+  - job_name: minio
+    metrics_path: /minio/v2/metrics/cluster
+    static_configs:
+      - targets: ['minio:9000']
```

And `platform/observability/**` is still in `deploy.yml`'s path filter at both the push trigger (line 9) and the `paths-filter` step (line 60). **Merging this fires an unattended production observability deploy** — that is intended, and it is why the PR is being held rather than merged unattended.

### Local checks

```
traefik-config.bats     28/28 pass
platform-services.bats  41/41 pass
deploy.bats             32/32 pass
validate.bats           17/17 pass
tests/checks (pytest)   29 passed
check_md_links.py       pass
check_destructive_commands.sh  pass
```

CI is green (`Run Tests` pass, Snyk pass). `mergeStateStatus` is `BLOCKED` on review only, not on conflict.

**Not merging — this one stays yours.**

jonhill90:
**Closing as superseded.** Platform MinIO shipped by a different route: #593, #594, #595, #597, #598, #599. `storage.hill90.com` now serves the platform MinIO (router `minio-console`, `tailscale-only@file` middleware, verified on the host), `app-minio` is stopped with `prod_app-minio-data` retained, and the platform container is healthy.

Established rather than assumed — every one of this PR's 24 files exists on `main`, and of its **588 substantive added lines (excluding the encrypted store), 571 are present verbatim** in `main`'s version of the same file. All **17** that are not are present in *better* form:

| This PR wanted | `main` has |
|---|---|
| `minio` in the backup target list | `db, app-db, minio, vault, infra, observability` — a superset; the line only differed because `app-db` was added later |
| inline `mc admin info` with `timeout 30` as the deploy readiness check | `MINIO_READY_TIMEOUT=5 … minio.sh status` — configurable timeout, and `minio.sh` carries the *same reasoning* in comments ("`mc ready` has no timeout and retries forever against an unreachable MinIO") plus a better failure message that distinguishes "MinIO answered and rejected the credentials, so this is not a readiness problem" from a genuine readiness timeout |
| `minio.sh` timeout/auth error handling | same, with a separate probe timeout and a 90s default |
| a comment warning that the Traefik `constraints` line accumulates one clause per local stack | the warning was heeded: line 93 records "`hill90-local-storage` left a conflict warning here. Both clauses kept", and the constraints line contains all seven clauses |

Checked specifically for the things this PR contributed that could plausibly have been lost:

- **Bucket policy** — present and more developed. `minio.sh` creates `admin`, `editor` and `viewer` policies idempotently via `mc admin policy create` (an upsert), matching the OIDC claim name.
- **OIDC detail** — present: `identity_openid`, `MINIO_IDENTITY_OPENID_*`, `MINIO_OIDC_CLIENT_SECRET`, and the `AssumeRoleWithWebIdentity` STS path are all on `main`, along with the `loginStrategy` finding about the console being removed from the AGPL build.
- **Tests** — `tests/scripts/platform-services.bats`, `tests/checks/test_deploy_scope.py`, `check_volume_names.py`, `check_env_surface.py` and `check_secrets_schema.py` are all on `main`.
- **Prometheus scrape** — `job_name: minio` at `minio:9000/minio/v2/metrics/cluster` is on `main`.

Nothing here is missing from `main`. The `MINIO_OIDC_CLIENT_SECRET` merge-ordering hazard in the description is also moot: the secret is in the store and `keycloak.sh apply` has run since.

Closed without merging; no code was taken from this branch.

PR #559, opened 2026-07-27
feat(observability): scrape the tenant's database, and say that constraints need a restart

jonhill90:
## Summary

Two requests from the app-arch lane, in one PR.

- **Prometheus scrapes the tenant's database** — a job for `app-postgres-exporter:9187`.
- **The Traefik constraint keeps both clauses**, and now says that changing that line needs a **restart, not a reload**.

> ### ⚠️ Stacked on #556 — merge after it
> Based on `feat/restore-minio`, not `main`, **on purpose**. Both files this touches are files #556 touches. Branching from `main` would have put #556 back into conflict — the bill that was just paid once, on the same `constraints` line. Until #556 merges, this PR's diff shows its commits too.

## Issue

- Requested by the app-arch lane. No Hill90 issue.

## Prometheus — the tenant database is unmonitored

The app keeps its own Postgres. Hill90's platform database is deliberately not shared, and its health check asserts *platform-only databases* — it is written to fail if app databases appear there. So the app's database needs its own exporter, and Hill90 scrapes it because Hill90 owns observability.

Reachability checked rather than assumed: Prometheus is on `edge` and `internal`, the app's exporter is on `internal`, so the name resolves once the app is up.

**Two things about this job are worth knowing before merging**, and both are written into the file rather than left to be discovered:

**1. The target does not exist yet — in either repo.** The app still ships `container_name: postgres-exporter`:

```
$ grep container_name hill90-app/deploy/compose/prod/docker-compose.db.yml
    container_name: postgres
    container_name: postgres-exporter
```

That collides with Hill90's platform exporter and cannot start beside it. So this job stays **DOWN** until the app both renames it and deploys — Phase A step 1 of the deployment runbook. The name `app-postgres-exporter` is forward-looking config.

**2. A down target fires a critical alert.** `alerts.yml` has:

```yaml
- alert: ServiceDown
  expr: up == 0
  for: 5m
  labels: {severity: critical}
```

There is **no Alertmanager** — no `alerting:` block in `prometheus.yml`, and no container locally or on the VPS — so nothing is paged. But it will read as a firing critical in Prometheus and Grafana until the app deploys, which could be a while.

I implemented it as asked, because the reasoning holds and a visible reminder of unmonitored tenant state beats silently forgetting the job later. But it should be a **choice rather than a surprise**, so: if you would rather not carry a firing critical, say so and I will hold the job until the app's rename lands.

This also corrects an earlier recommendation of mine. The tenancy assessment (#558) said the app should *delete* its `postgres-exporter` as a duplicate of an infra concern. That was wrong — it followed from Hill90 owning observability, but missed that Hill90's exporter points at Hill90's Postgres only. Given the app keeps its own database, the exporter should be **renamed, not deleted**.

## Traefik — restart, not reload

The constraint line already carries both contested clauses on this base, seven in total, so nothing is reintroduced or lost:

```
hill90-local-edge  hill90-local-observability  hill90-local-identity
hill90-local-platform  hill90-local-db  hill90-local-storage  hill90-local
```

What is new is the note above it. That line is **static** configuration, and only the `file` provider's dynamic directory is watched (`watch: true`). Editing constraints and waiting appears to do nothing: the routers stay dropped, no error is logged, and **the file on disk already shows the value you wanted**. That is the most misleading state this file can be in, and it is the same *present, plausible and inert* family tracked in #557 — with the twist that here the config is correct and only the process is stale.

## Validation Evidence

**Infra —** both files parse, and the assertion that matters is explicit:

```
scrape jobs: prometheus traefik keycloak postgres-exporter node-exporter
             cadvisor grafana loki tempo minio app-postgres-exporter
constraint clauses: 7
both contested clauses present: OK
```

**Checks —** all local suites pass:

```
traefik-config.bats     28/28      tests/checks (pytest)   29 passed
platform-services.bats  41/41      check_env_surface.py    38 vars, consistent
validate.bats           17/17      check_secrets_schema.py all checks passed
deploy.bats             32/32      check_md_links.py       pass
```

## Notes

**This PR fires an unattended production deploy on merge.** `platform/observability/**` is in `deploy.yml`'s path filter at both the push trigger and the `paths-filter` step. Same property that is holding #556. Merge when someone is awake and watching.

## Test plan

- [x] Both YAML files parse; job list and clause count asserted
- [x] Prometheus/exporter network reachability confirmed (`edge`+`internal` / `internal`)
- [x] All local suites pass
- [ ] CI checks pass

**Do not merge** — stacked behind #556, and both are Jon's.

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -149,6 +149,12 @@ MINIO_ROOT_PASSWORD=localdev-minio-password
 # the AGPL build (removed May 2025), so this does not give you a login button.
 MINIO_OIDC_CLIENT_SECRET=local-minio-secret-0123456789
 
+# Routing. Production resolves to storage.hill90.com with no environment set.
+MINIO_HOST=storage
+MINIO_PUBLIC_URL=http://storage.localtest.me:8080
+# The token claim MinIO reads the policy name from. keycloak.sh emits it.
+MINIO_OIDC_CLAIM_NAME=minio_policy
+
 # The local Portainer admin. portainer.sh uses it to call the Portainer API, and
 # it is your way in when Keycloak is unavailable. Portainer 2.39 requires the
 # admin to be created within a few minutes of first start, using a setup token
--- a/.env.local.example
+++ b/.env.local.example
@@ -140,6 +140,15 @@ GRAFANA_OIDC_ENABLED=true
 GRAFANA_OIDC_CLIENT_SECRET=local-grafana-secret-0123456789
 PORTAINER_OIDC_CLIENT_SECRET=local-portainer-secret-0123456789
 
+# --- Object store (MinIO) ----------------------------------------------------
+# Root credentials for the local object store. Disposable; production keeps its
+# own in SOPS.
+MINIO_ROOT_USER=minioadmin
+MINIO_ROOT_PASSWORD=localdev-minio-password
+# Gates the S3/STS path against Keycloak. MinIO's console has NO SSO login in
+# the AGPL build (removed May 2025), so this does not give you a login button.
+MINIO_OIDC_CLIENT_SECRET=local-minio-secret-0123456789
+
 # The local Portainer admin. portainer.sh uses it to call the Portainer API, and
 # it is your way in when Keycloak is unavailable. Portainer 2.39 requires the
 # admin to be created within a few minutes of first start, using a setup token
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,15 @@ jobs:
         run: bats tests/scripts/
 
       - name: Validate compose files
+        # Credentials that must never fall back to a default use ${VAR:?...},
+        # so `config` refuses to render without them. That guard is deliberate:
+        # an empty MinIO root password silently starts the server on
+        # minioadmin/minioadmin with only a WARN. Supply throwaway values here —
+        # this step validates STRUCTURE, not secrets.
+        env:
+          MINIO_ROOT_USER: placeholder
+          MINIO_ROOT_PASSWORD: placeholder
+          MINIO_OIDC_CLIENT_SECRET: placeholder
         run: |
           for f in deploy/compose/prod/docker-compose*.yml; do
             echo "Validating $f..."
--- a/.github/workflows/deploy-minio.yml
+++ b/.github/workflows/deploy-minio.yml
@@ -1,0 +1,18 @@
+name: Manual Deploy Object Store (Prod)
+
+on:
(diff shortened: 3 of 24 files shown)


jonhill90:
**Closing.** The first half is dead on its premise; the second half is still true and worth keeping, and is called out below so it is not lost with the branch.

This PR was also stacked on `feat/restore-minio` (#556), which is now closed as superseded — so its base is gone regardless.

## Half 1 — scrape the tenant's database: obsolete

The PR's stated premise was "**The app keeps its own Postgres.** Hill90's platform database is deliberately not shared, and its health check asserts *platform-only databases*." That is no longer true.

`app-postgres` was **retired on 2026-07-31**. The tenant's three databases — `hill90_api`, `hill90_akm`, `hill90_litellm` — now live on the **platform** instance, owned by the NOSUPERUSER role `hill90_app`, and the platform-only-databases boundary was revisited deliberately rather than worked around.

The target this PR adds a scrape job for does not exist and now cannot:

- `app-postgres-exporter` appears in **0 files** in this repository and **0 files** in hill90-app
- the only exporter container on the host is the platform's own `postgres-exporter`

The PR itself noted "the target does not exist yet — in either repo", expecting the app to rename its container. Instead the app stopped having a database.

**The underlying concern does partly survive, in a different shape**, and is already recorded in the published docs rather than needing this PR. Hill90's exporter connects to `postgres:5432/hill90`. Cluster-wide views are visible from any database, so `pg_stat_database` for the three tenant databases **is** now scraped. Per-table views are not — `pg_stat_user_tables` is per-database — so table and index statistics for the tenant's databases remain unscraped. Closing that gap needs **multi-target scraping** (`postgres-exporter` v0.17.1 supports `/probe?target=`), not a second exporter, so none of this PR's diff would help.

## Half 2 — worth extracting, not on `main`

The Traefik `constraints` clauses are intact on `main` (all seven, with a comment recording that the collision was resolved by keeping both). But **`main` has no mention of restart-versus-reload** — grep for `restart|reload` in `platform/edge/traefik.local.yml` returns nothing.

This PR adds it:

```
# CHANGING THIS LINE REQUIRES A TRAEFIK RESTART, NOT A RELOAD. It lives in
# ... the most misleading state this file can be in. Restart the container:
#   docker restart hill90dev-traefik
```

That is correct — `providers.docker.constraints` is static configuration, so Traefik will not pick it up on a reload — and it applies to a line that has already caused two merge collisions. It is four lines of comment on a file that still exists, so it should be lifted into its own change rather than kept alive on a stacked branch with a dead base.

Closed without merging.

PR #633, opened 2026-08-01
docs: CLAUDE.md contradicted itself about two of the three retirements

jonhill90:
Both claims were **verified against the host before touching either**, and both are false.

## 1. The app *has* been cut over

The Postgres section read:

> *"The app has **NOT** been cut over. It still reads and writes `app-postgres`. The
> databases on this platform are empty and waiting."*

…while the section forty lines below already recorded `app-postgres` as retired on
2026-07-31. **A cold session reads the first and stops.**

`Verified 2026-07-31 12:05 UTC` on the host:

```
app-api DATABASE_URL = postgresql://hill90_app:<redacted>@postgres:5432/hill90_api
app-api resolves 'postgres' -> 172.19.0.9          (this platform's Postgres)
docker ps -a --filter name=app-postgres            -> nothing. Not even stopped.
databases on the instance: hill90, hill90_akm, hill90_api, hill90_litellm
```

All three clauses were wrong. `app-postgres-cutover-plan.md` is a record of a plan that
was **carried out**, not of pending work.

## 2. `app-auth` is gone, not "still running"

The fast-facts entry said `app-auth.hill90.com` *"is still running"* and that retiring it
*"is pending local parity, not done"* — contradicted by the same file's own record of
`app-keycloak` going on 2026-07-30.

`Verified 2026-07-31 12:05 UTC`: **no `app-keycloak` container exists**, and the hostname
returns **404**.

## Swept the rest of the file

You asked me to re-read for anything else the three retirements falsified, so I did rather
than fixing only the reported paragraph. **Three passages that look stale are correct**,
recorded here so the next reader does not re-open them:

| Passage | Why it stands |
|---|---|
| the 2026-07-30 row counts *"on `app-postgres`"* | a **dated historical measurement** of a container that no longer exists — which is exactly what dating a claim is for |
| *"the tenant's local stack still points at its own `app-keycloak`"* | about **local**, which deliberately still runs the app's own services |
| *"this platform held at exactly 13"* | the **2026-07-29 yank-out test**, dated; the paragraph above already explains why today's baseline is 16 |

`app-minio`'s entry is also still accurate — stopped, not removed, window open until
2026-08-01 01:41 UTC.

## Verification

`check_md_links` passes. Documentation only; **nothing deployed and nothing on the host
touched** — every check above was read-only.

*(PR 2, for the realm rebuild trap, follows separately. It found a third missing client the
brief did not mention.)*

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,10 +155,20 @@ revisited deliberately rather than worked around.
 > See
 > [tenant-databases-on-platform-postgres.md](docs/decisions/tenant-databases-on-platform-postgres.md).
 >
-> **The app has NOT been cut over.** It still reads and writes `app-postgres`. The
-> databases on this platform are empty and waiting. The change set that repoints it
-> is written and **not applied** —
-> [app-postgres-cutover-plan.md](docs/decisions/app-postgres-cutover-plan.md).
+> **The cutover HAPPENED on 2026-07-31, and this paragraph said the opposite until
+> it was checked.** It read *"The app has NOT been cut over. It still reads and writes
+> `app-postgres`. The databases on this platform are empty and waiting."* All three
+> clauses are now false, and the section immediately below already recorded the
+> retirement — so this file contradicted itself.
+>
+> `Verified 2026-07-31 12:05 UTC` on the host: `app-api`'s `DATABASE_URL` is
+> `postgresql://hill90_app:…@postgres:5432/hill90_api` — this platform's Postgres,
+> which the container resolves to `172.19.0.9` — and **no `app-postgres` container
+> exists at all**, not even stopped. The instance holds `hill90`, `hill90_akm`,
+> `hill90_api` and `hill90_litellm`.
+>
+> [app-postgres-cutover-plan.md](docs/decisions/app-postgres-cutover-plan.md) is
+> therefore a record of a plan that was carried out, not of pending work.
 
 **This is greenfield, not a migration — with one qualifier that matters.**
 hill90-app reached the VPS for the first time on 2026-07-29. Export, import,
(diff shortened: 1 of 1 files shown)


jonhill90:
Holding this — the two corrections are right and host-verified, but the replacement paragraph introduces a new inaccurate claim of exactly the kind this PR exists to remove.

It says: *"The instance holds `hill90`, `hill90_akm`, `hill90_api` and `hill90_litellm`."*

Measured on the host just now, as the platform role (`-U postgres` fails outright — `role "postgres" does not exist`, so an empty result from that instrument is blindness, not absence):

```
$ docker exec postgres psql -U hill90 -d postgres -tAc \
    "select datname from pg_database where not datistemplate order by 1"
hill90 hill90_akm hill90_api hill90_litellm keycloak postgres
```

Six, not four. The omitted two are `keycloak` — load-bearing, it is the platform Keycloak’s own store — and `postgres`. In a paragraph correcting "the databases on this platform are empty and waiting", that list reads as the enumeration, and CLAUDE.md elsewhere makes an explicit *"that is the complete list of databases on this instance"* claim which this would now contradict. Same shape as the earlier `grafana`-read-as-a-database error.

Fix: state all six, or say plainly that you are naming only the tenant-owned subset and which role owns it. Everything else in the diff stands — I re-verified `app-postgres` absent, `app-keycloak` absent, `app-auth.hill90.com` 404, and `app-api` pointed at `postgres:5432/hill90_api`. Push the correction and I will merge.

jonhill90:
Corrected and pushed to the same branch — thank you, that was a real defect and of exactly the class this PR exists to remove.

## What went wrong, precisely

Not an omission. **My own filter, forgotten.** The query behind the sentence was:

```sql
select datname from pg_database where datname LIKE 'hill90%'
```

I chose that `WHERE` clause, then wrote the filtered result up as an enumeration of the instance. Nothing was hidden from me — I forgot my own predicate between running it and quoting it. Re-measured unfiltered as the platform role:

```
hill90  hill90_akm  hill90_api  hill90_litellm  keycloak  postgres
```

Six. The two I dropped were `postgres` and — the one that matters — `keycloak`, the platform Keycloak's own store.

## The fix: name the subset, do not re-enumerate

You offered both options; I took the second, because the first would put a second enumeration in a file that already has one fourteen lines above:

> Role `hill90_app` (`NOSUPERUSER`) owns `hill90_akm`, `hill90_api` and `hill90_litellm`; `hill90`, `keycloak`, `postgres` and the templates stay with the platform role. **That is the complete list of databases on this instance.**

That claim is accurate and I verified it independently — ownership included:

```
hill90_akm, hill90_api, hill90_litellm  ->  hill90_app
hill90, keycloak, postgres              ->  hill90
```

So the paragraph now names **the three databases the tenant role `hill90_app` owns**, says plainly that this is the tenant-owned subset and *not* the instance list, and points upward for the complete one. Two enumerations in one file is precisely how they drift apart.

## The instrument trap is now in the file

Recorded where the next person will reach for it: `psql -U postgres` fails on this host with `role "postgres" does not exist`, so an empty result from that invocation is **blindness, not absence**. Use `-U hill90`, and do not filter.

That is the fourth time today this class has caught me, and the first time it caught me *inside a correction to the same class* — which is a better argument for the rule in `CONTRIBUTING.md` than the rule itself makes.

CI green, nothing else in the diff touched, not merged.

*(Noted on #634 — thank you for re-running the positive control yourself rather than taking the transcript's word for it.)*

PR #656, opened 2026-08-03
fix(vault): the AppRole authorisation gap was documented, not fixed — close it

jonhill90:
#653 shipped a decision record and a check. It changed **no policy and no `vault.sh`**, so the gap it described was still live.

**Answering the question the PR title did not: NOT FIXED.**

## Measured on the live vault, by reading the paths

| Role | Path | Result |
|---|---|---|
| `db` | `secret/shared/database` | **403 denied** |
| `auth` | `secret/shared/database` | **403 denied** |
| `auth` | `secret/auth/config` | **403 denied** |
| `infra` | `secret/infra/traefik` | read ok |
| `observability` | `secret/observability/grafana` | read ok |

All four roles **log in**. Two cannot read anything — the outage's exact shape: authenticating is not being authorised.

## Root cause, in code

`cmd_setup` binds every service to `token_policies=policy-<svc>` for `VAULT_SERVICES="db auth infra observability"`, while `cmd_policy_apply` writes only the `.hcl` files that exist. There was **no `policy-auth.hcl` and no `policy-db.hcl`** — those roles carried a policy that had never been written. `infra` and `observability` were fine because their files existed. **The reseed did not single auth out; the policy directory did.**

## Fixed here

```
+ policy-db.hcl     secret/{data,metadata}/shared/database
+ policy-auth.hcl   secret/{data,metadata}/shared/database, .../auth/config
```

**And the over-grant direction**, which the brief was right to weight equally: `policy-infra` granted `secret/data/infra/*` while infra declares one path, and `policy-observability` was the same shape. Both are now exactly their declared paths. An over-grant fails no deploy and appears in no outage, which is precisely why it survives.

## Two instrument failures, both mine

**1.** My first probe used `bao kv get` and reported **all four roles denied**. Wrong. `bao kv get` first calls `sys/internal/ui/mounts/<path>` to detect the KV version, and no narrow policy grants that endpoint — so it 403s for every narrow role regardless of whether the data path is readable. `bao read secret/data/<path>` skips mount discovery. With the correct probe, infra and observability read fine and the gap is exactly the two documented roles.

**2.** I ran the reads with `2>/dev/null`, which hid the 403 text and made a denial look like an empty secret.

The first would have had me report a four-role catastrophe. Both are recorded in the script.

## The new check

`scripts/checks/vault-approle-real-read-test.sh` **performs the read** rather than asking the capability API — the outage surfaced *as* a capability check returning 403, so a check asking the same question the broken preflight asked cannot independently confirm the fix. It derives `VAULT_SERVICES` from `vault.sh` so it cannot cover fewer roles than the setup creates, passes `role_id`/`secret_id` through the environment rather than argv (#607), and never emits a secret value.

**Confirmed red before trusted green:** against the live broken vault it exits 1 and names the three failing reads — the table above *is* its output.

## What this PR does not do

**It does not make the live vault correct.** Applying policies needs a privileged token; the root token is revoked by design and deploys are pipeline-only. The policy files land here and `vault.sh policy-apply` must run through the pipeline. Until then `db` and `auth` still cannot read, and the estate keeps working only because `vault_load_secrets` falls back to SOPS.

## Guards and baseline

#651/#652 guards intact — deploy-time resolution untouched, full bats suite **382 passing, 0 failing**. Baseline: 16 platform containers by name, **0 unhealthy, 0 restarting**. No secret value printed.

### DIFF
--- a/deploy/compose/prod/docker-compose.minio.yml
+++ b/deploy/compose/prod/docker-compose.minio.yml
@@ -77,6 +77,22 @@ services:
       # MinIO reads the policy to grant from this claim. keycloak.sh adds a
       # mapper that emits it; a token without it gets no access rather than
       # broad access.
+      #
+      # THE EMPTY CASE WAS NEVER THE PROBLEM. The mapper is a REALM-ROLE mapper,
+      # so this claim arrives populated with role names, not a policy name —
+      # measured 2026-08-03 on RELEASE.2025-09-07T16-13-09Z:
+      #
+      #   values naming no policy   -> ignored
+      #   nothing matches at all    -> STS refuses, "None of the given policies
+      #                                (...) are defined"
+      #   several match             -> the grant is their UNION
+      #
+      # The union rule is the sharp edge: every token also carries
+      # offline_access, uma_authorization and default-roles-platform, which every
+      # realm account holds. A MinIO policy named after one of those is granted
+      # to everybody who can log in. scripts/checks/minio-policy-names-test.sh
+      # fails on exactly that collision. See
+      # docs/decisions/minio-stage3-2026-08-03.md.
       - MINIO_IDENTITY_OPENID_CLAIM_NAME=${MINIO_OIDC_CLAIM_NAME:-minio_policy}
       - MINIO_IDENTITY_OPENID_REDIRECT_URI=${MINIO_PUBLIC_URL:-https://storage.hill90.com}/oauth_callback
     healthcheck:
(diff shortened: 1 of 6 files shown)


jonhill90:
Closing: this branch ended up pointing at 67548e14, which is another lane's MinIO Stage 3 commit and belongs to #655. My AppRole policy work was uncommitted when a concurrent checkout replaced the working tree, so this PR carried the wrong changes under the wrong title. #655 is unaffected. Reopening the real change on a fresh branch.

PR #657, opened 2026-08-03
fix(vault): the AppRole authorisation gap was documented, not fixed — close it

jonhill90:
#653 shipped a decision record and a check. It changed **no policy and no `vault.sh`**, so the gap it described was still live.

**Answering the question the title did not: NOT FIXED.**

## Measured on the live vault, by reading the paths

| Role | Path | Result |
|---|---|---|
| `db` | `secret/shared/database` | **403 denied** |
| `auth` | `secret/shared/database` | **403 denied** |
| `auth` | `secret/auth/config` | **403 denied** |
| `infra` | `secret/infra/traefik` | read ok |
| `observability` | `secret/observability/grafana` | read ok |

All four roles **log in**. Two cannot read anything — the outage's exact shape: authenticating is not being authorised.

## Root cause, in code

`cmd_setup` binds every service to `token_policies=policy-<svc>` for `VAULT_SERVICES="db auth infra observability"`, while `cmd_policy_apply` writes only the `.hcl` files that exist. There was **no `policy-auth.hcl` and no `policy-db.hcl`**. `infra` and `observability` were fine because their files existed — **the reseed did not single auth out; the policy directory did.**

```
+ policy-db.hcl     secret/{data,metadata}/shared/database
+ policy-auth.hcl   secret/{data,metadata}/shared/database, .../auth/config
```

**The over-grant direction, weighted equally as asked:** `policy-infra` granted `secret/data/infra/*` while infra declares one path; `policy-observability` was the same shape. Both are now exactly their declared paths. An over-grant fails no deploy and appears in no outage, which is why it survives.

## Two instrument failures, both mine

**1.** My first probe used `bao kv get` and reported **all four roles denied** — wrong. `kv get` first calls `sys/internal/ui/mounts/<path>` to detect the KV version, and no narrow policy grants that endpoint, so it 403s for every narrow role regardless of whether the data path is readable. `bao read secret/data/<p>` skips mount discovery. The difference between a four-role catastrophe and the truth.

**2.** I ran the reads with `2>/dev/null`, hiding the 403 text and making a denial look like an empty secret.

## The new check

`scripts/checks/vault-approle-real-read-test.sh` **performs the read** rather than asking the capability API — the outage surfaced *as* a capability check returning 403, so a check asking the same question the broken preflight asked cannot independently confirm the fix. It derives `VAULT_SERVICES` from `vault.sh`, passes `role_id`/`secret_id` through the environment rather than argv (#607), and never emits a secret value.

**Confirmed red before trusted green** — the table above *is* its output against the live vault.

## What this does not do

**It does not make the live vault correct.** Applying policies needs a privileged token; the root token is revoked by design and deploys are pipeline-only, so `vault.sh policy-apply` must run through the pipeline. Until then `db` and `auth` still cannot read, and the estate works only because `vault_load_secrets` falls back to SOPS.

## Guards and baseline

#651/#652 intact — deploy-time resolution untouched, bats **382 passing, 0 failing**. Baseline: 16 platform by name, **0 unhealthy, 0 restarting**. No secret value printed.

> Recreated on a clean branch: a concurrent checkout in this shared worktree replaced my uncommitted files and left the previous branch pointing at another lane's commit. That PR (#656) was closed as mislabelled; **#655 and its branch are untouched.**

### DIFF
--- a/deploy/compose/prod/docker-compose.minio.yml
+++ b/deploy/compose/prod/docker-compose.minio.yml
@@ -77,6 +77,22 @@ services:
       # MinIO reads the policy to grant from this claim. keycloak.sh adds a
       # mapper that emits it; a token without it gets no access rather than
       # broad access.
+      #
+      # THE EMPTY CASE WAS NEVER THE PROBLEM. The mapper is a REALM-ROLE mapper,
+      # so this claim arrives populated with role names, not a policy name —
+      # measured 2026-08-03 on RELEASE.2025-09-07T16-13-09Z:
+      #
+      #   values naming no policy   -> ignored
+      #   nothing matches at all    -> STS refuses, "None of the given policies
+      #                                (...) are defined"
+      #   several match             -> the grant is their UNION
+      #
+      # The union rule is the sharp edge: every token also carries
+      # offline_access, uma_authorization and default-roles-platform, which every
+      # realm account holds. A MinIO policy named after one of those is granted
+      # to everybody who can log in. scripts/checks/minio-policy-names-test.sh
+      # fails on exactly that collision. See
+      # docs/decisions/minio-stage3-2026-08-03.md.
       - MINIO_IDENTITY_OPENID_CLAIM_NAME=${MINIO_OIDC_CLAIM_NAME:-minio_policy}
       - MINIO_IDENTITY_OPENID_REDIRECT_URI=${MINIO_PUBLIC_URL:-https://storage.hill90.com}/oauth_callback
     healthcheck:
--- a/docs/decisions/minio-stage3-2026-08-03.md
+++ b/docs/decisions/minio-stage3-2026-08-03.md
@@ -1,0 +1,170 @@
+# MinIO Stage 3: what MinIO does with a multi-valued claim, and the remedy that follows
+
+`Measured 2026-08-03 against MinIO RELEASE.2025-09-07T16-13-09Z — the deployed image.`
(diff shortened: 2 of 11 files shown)


jonhill90:
**Closed automatically because the branch became empty on rebase — the content is already in `main`.**

Not because it was abandoned. Rebasing onto `main` produced an add/add conflict on `platform/vault/policies/policy-db.hcl`, and the reason is that **every file in this PR had already landed in `main` via #655**. The concurrent checkout I flagged when opening this PR did more than move a branch ref: it swept my uncommitted files into another lane's commit, so the AppRole policy work merged under a MinIO title.

Diffed file by file against `main` before resolving. The only delta in the entire PR was **one reworded comment line** in `policy-db.hcl`; `policy-auth.hcl`, `policy-infra.hcl`, `policy-observability.hcl` and `scripts/checks/vault-approle-real-read-test.sh` were byte-identical. I resolved to `main`'s wording, the commit emptied, and GitHub closed the PR.

`#655`'s change to `vault_load_secrets` — no declared paths returns **non-zero**, not 0 — is intact and was never touched by this branch. bats: **382 passing, 0 failing**.

**The live vault is still broken**, re-measured after the rebase:

```
db             secret/shared/database             403-denied
auth           secret/shared/database             403-denied
auth           secret/auth/config                 403-denied
infra          secret/infra/traefik               read-ok
observability  secret/observability/grafana       read-ok
```

Merging the policy files does not apply them. `vault.sh policy-apply` still has to run through the pipeline.

PR #716, opened 2026-08-04
fix(ci): watchdog's own cleanup failure masked a real delivery

jonhill90:
Follow-up to #715. The second live run against the real, still-failing `Tenant Baseline Agreement` workflow showed the alert POST to Alertmanager succeeding, then the cleanup `docker exec alertmanager rm -f` failing with permission denied (docker cp writes as root; the container execs as uid 65534) — which, under `set -euo pipefail`, reported the whole step, and the job, as failed on a cleanup command rather than the actual delivery. Made cleanup best-effort, and added a step that checks Alertmanager's own container log for a notify/smtp/email line — the strongest signal available without mailbox access, explicitly not claimed as inbox confirmation.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/scheduled-checks-watchdog.yml
+++ b/.github/workflows/scheduled-checks-watchdog.yml
@@ -162,6 +162,26 @@ jobs:
                 --header="Content-Type: application/json" \
                 --post-file=/tmp/alert.json \
                 http://localhost:9093/api/v2/alerts
-              docker exec alertmanager rm -f /tmp/alert.json
+              # Best-effort only. The container runs as uid 65534 (see the
+              # "DELIBERATELY NOT PUBLIC" comment on the alertmanager
+              # service), and `docker cp` writes as root regardless of that
+              # — so this non-root exec cannot always remove what cp wrote.
+              # A leftover /tmp/alert.json is a stray file in a container
+              # that already gets torn down and rebuilt on every deploy; it
+              # must never be the reason the POST above is reported as
+              # failed.
+              docker exec alertmanager rm -f /tmp/alert.json || true
               rm -f /tmp/scheduled-workflow-alert.json
             '
+
+      # Alertmanagers API accepting the POST is a handoff, not proof an
+      # email arrived — the same boundary the SMTP precedent this reuses
+      # only proved as far as "AUTH accepted" (see alertmanager.yml.tmpl).
+      # This step names exactly what it can see: whether Alertmanager's own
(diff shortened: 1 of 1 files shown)


jonhill90:
## Confirmed live, tonight — and why this is more than a cleanup annoyance

Dispatched `Tenant Baseline Agreement` twice tonight (once directly, once by coincidence of a near-simultaneous second dispatch) against its genuine, still-unresolved red cause (`TENANT_BASELINE_TOKEN` missing). Both triggered `Scheduled Checks Watchdog` independently. Both runs show the identical signature this PR fixes: the SSH payload's commands execute strictly in sequence under `set -euo pipefail` — `docker cp`, then the `wget` POST to `/api/v2/alerts`, then `docker exec alertmanager rm -f /tmp/alert.json`. Both runs reached and failed on the `rm` line (`Operation not permitted`), which under `set -e` is only reachable if the two commands before it — the copy and the POST — both exited 0. So on both runs, delivery to Alertmanager's API genuinely succeeded, and the job still reported red.

Read from the host directly (not inferred): `alertmanager_notifications_total{integration="email"}` is 11 and the sum of `alertmanager_notifications_failed_total{integration="email"}` across every reason is 0. Alertmanager routed the alert to the `email` receiver and handed it to the mail server without a single recorded failure. What that does not establish, and what this PR's own new step is careful not to claim either: nobody has seen it in an inbox. That boundary — handoff to SMTP proven, arrival not — is real and stays where it is; this PR's notify-log step is honest about not crossing it.

**Say plainly why the exit-1-on-`rm` shape matters more than an untidy exit code.** This is the same family of defect this repo spent today naming and fixing in the other direction — an operation that reports success while quietly failing. This is that shape inverted: **the watchdog succeeds and reports failure.** A watcher whose job is "tell a human when something red needs attention" that is *itself* permanently red — because of its own unrelated cleanup step, not because of what it's watching — trains every reader to expect its red and stop looking closely at it. The first time a genuinely new, different failure shows up in this same job, it will look exactly like the noise everyone has already learned to skip past. That is a worse failure mode than a stray file in a container that gets rebuilt on every deploy, and it's the reason this isn't just tidiness.

Confirms #716's fix is aimed at the right thing: `rm -f ... || true` stops an unrelated, expected-sometimes-to-fail cleanup command from being able to flip the job's own conclusion, and the added notify-log step gives a real (bounded, explicitly-not-inbox) signal about what happened after the handoff. Not implementing anything further tonight — this stands as filed.

PR #717, opened 2026-08-04
diag: throwaway read-only IPv6 probe for #542 (delete after one dispatch)

jonhill90:
Read-only diagnostic to answer #542 directly from the host rather than by reasoning from compose files: is anything listening on IPv6, is Traefik bound to it, and could an inbound client reach anything over v6 if a AAAA record existed.

No file written, no service touched, no DNS changed. Reuses the exact SSH/Tailscale connection pattern already trusted in `deploy-drift.yml` and `scheduled-checks-watchdog.yml` — same secrets, same access, nothing new granted.

Requesting merge + a `workflow_dispatch` run, then this branch and workflow file should be deleted — not meant to be kept.

### DIFF
--- a/.github/workflows/diag-ipv6-readonly.yml
+++ b/.github/workflows/diag-ipv6-readonly.yml
@@ -1,0 +1,103 @@
+name: Diagnostic — IPv6 reachability read-only (throwaway, #542)
+
+# THROWAWAY. Exists only to answer #542's question by reading the host, not by
+# reasoning from compose files: is anything listening on IPv6, is Traefik bound
+# to it, and could an inbound client reach anything over v6 if a AAAA record
+# existed. Read-only: no file is written, no service is restarted, no DNS is
+# touched. This workflow is not meant to be kept — delete the branch after the
+# one dispatch that answers the question.
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  probe:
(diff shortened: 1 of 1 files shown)


jonhill90:
Closing unmerged — the question is answered, and this workflow should not exist.

I ran the probe directly over the existing SSH path instead, which needed no repository change at all:

- `ip -6 addr show scope global` → **zero global IPv6 addresses on the host.**
- `ss -6 -ltnp` → **no IPv6 listeners at all.**
- `docker port traefik` → Docker publishes `[::]:80` and `[::]:443`, which is the *daemon's* dual-stack bind on the loopback-only v6 stack, not reachability from outside.

So #542's premise does not hold as written: the VPS does **not** have a usable public IPv6 address today. The missing AAAA record is not a gap — it is the correct state, because there is nothing for it to point at. That moves #542 from "latent" to "harmless, and the issue's premise is stale".

**Why this PR is closed rather than merged and then deleted.** A workflow committed with "delete after one dispatch" in its own title is the shape this repo has removed twice today and filed [#712](https://github.com/jonhill90/Hill90/issues/712) about: a check that runs only when someone remembers, living in the tree where the next reader assumes it is maintained. If the answer is needed once, run it once from a shell. If it is needed repeatedly, it deserves a schedule and a reason — not a delete-me note.

No changes were made to DNS, the host, or any service.

PR #722, opened 2026-08-05
fix(vault): sync workflow authenticates via AppRole, not a periodic token (#711)

jonhill90:
## Summary

Closes the root cause #711 diagnosed: the weekly `vault-sync-to-sops` workflow held a periodic token (`period=768h`) that renewed itself via `bao token renew` each run. Renewal silently capped at OpenBao's `config.hcl` `max_lease_ttl=24h` instead of the token's own period, because periodic-token renewal only exceeds the system ceiling when the renewing identity holds `sudo` on `auth/token/renew-self` — which `policy-sync` never granted. The token was dead most weeks.

**Recommendation and decision are recorded on #711's comment thread** — four options plus a fifth from the issue's own diagnosis were argued there before any code was written, per instruction. Summary: AppRole removes the renewal/ceiling mismatch as a category (a fresh login every run either succeeds inside the 24h ceiling or fails immediately — no multi-week gap for this bug class to hide in again), and it avoids granting `sudo` anywhere, which appears in exactly one policy in this estate today (`policy-admin`, break-glass) and would otherwise become the first non-break-glass `sudo` grant on an automated, CI-triggered credential.

## Precondition checked before writing this, not assumed

Per instruction, ran `scripts/checks/vault-approle-login-test.sh` **on the VPS**, not this workstation — a workstation attempt correctly reported "Container openbao is not running" (no such container locally), which is the exact blindness trap the estate has written down before. On the VPS: the four maintained AppRoles (`db`, `auth`, `infra`, `observability`) pass both a real login and a real read of their declared paths. Full output and the discovery of five separate, already-dead orphaned AppRole pairs (`ai`/`api`/`mcp`/`minio`/`ui`) are on #711's thread; that cleanup is filed separately as **#721**, out of scope here.

## What changed

- **`scripts/vault.sh`**: `"sync"` added to `VAULT_SERVICES` — it gets the same AppRole shape as the other four via the existing `cmd_setup` loop: `token_ttl=1h`/`token_max_ttl=4h`, comfortably inside the 24h ceiling. New `cmd_setup_sync_approle` mints and stores just the `sync` credential (`VAULT_SYNC_ROLE_ID`/`VAULT_SYNC_SECRET_ID`), deliberately **not** `cmd_bootstrap_approles`, which would also mint fresh (harmless but noisy) secret-ids for the other four services. `cmd_setup_sync_token` is left in place, unmodified — `vault-init.yml` and `vault-reinitialize.yml` still call it as part of full-vault-bootstrap automation; retiring it there is a separate, larger change than this issue asked for.
- **`.github/workflows/vault-sync-to-sops.yml`**: "Renew sync token" is gone — there's nothing to renew. Replaced by an AppRole login step that mints a fresh token every run. Credentials and the resulting token travel over stdin/env, never argv, matching this file's existing security discipline (documented in the step's own comment, not just here).
- **`scripts/checks/vault-approle-login-test.sh`**: now derives its role list from `scripts/vault.sh`'s own `VAULT_SERVICES` instead of a separately hardcoded default — fixes a real two-lists-that-can-drift gap this issue's own precondition check surfaced (this script silently only checked "the four platform roles" while `vault-approle-real-read-test.sh` already read the canonical list to avoid exactly that).
- **`scripts/_common.sh`**: `vault_paths_for_service` gets a `sync` case (reusing the already-seeded `secret/shared/database`) so the real-read test exercises a genuine read for the new role too, not just a login.
- **`tests/scripts/vault.bats`**: new tests, plus one existing test fixed for real — `"vault-sync-to-sops workflow reads VAULT_SYNC_TOKEN from SOPS"` was asserting the bare string appears anywhere in the file, which kept passing against the new version purely because of this PR's own explanatory comments mentioning the old name by history. Rewritten to assert on the functional `VAULT_SYNC_TOKEN=` read specifically (now absent) and the new `VAULT_SYNC_ROLE_ID=`/`VAULT_SYNC_SECRET_ID=` reads (now present) — the same distinction the whole rest of this exercise has been about.

## The one-time manual step — unavoidable under any option, stated exactly

A human must run `vault.sh setup-sync-approle` (with `BAO_TOKEN` set) on a host that can reach OpenBao, then commit and push the resulting SOPS change, **before** the schedule can recover on its own. No config change resurrects the already-dead `VAULT_SYNC_TOKEN` — a token past its own TTL doesn't come back under any of the options that were on the table.

**The first scheduled run after this merges, and before that step runs, is EXPECTED to fail.** The workflow's own error message says exactly that (`"Run 'vault.sh setup-sync-approle' first (h#711) — this is expected to fail on the first scheduled run after that migration lands, until the one-time credential mint has happened and been pushed."`) rather than reading like a regression to whoever's on call that week.

## Positive control — captured live, real output

Read-only against the running OpenBao on the VPS (a failed auth attempt mutates nothing):

```
$ ROLE_ID="00000000-0000-0000-0000-000000000000" SECRET_ID="00000000-0000-0000-0000-000000000000" \
  docker exec -e BAO_ADDR=http://127.0.0.1:8200 -e ROLE_ID -e SECRET_ID openbao \
  sh -c 'bao write -field=token auth/approle/login role_id="$ROLE_ID" secret_id="$SECRET_ID"'; echo "EXIT:$?"

Error writing data to auth/approle/login: Error making API request.
URL: PUT http://127.0.0.1:8200/v1/auth/approle/login
Code: 400. Errors:
* invalid role or secret ID
EXIT:2
```

Same shape #721's five already-dead orphaned credentials produce independently — a wrong or stale credential fails loudly and immediately, every time, through the exact chain the workflow now uses. Verified the success path too, through the identical mechanism, with a real working credential (`db`) returning a real token (not reproduced here).

## Test plan

- [x] `bash -n` on every modified script: clean.
- [x] `shellcheck -S warning`: 0 new warnings on `scripts/vault.sh` or `scripts/_common.sh` (diffed against the pre-change versions to confirm — 0 before, 0 after, both files).
- [x] `python3 scripts/checks/check_declared_paths_are_seeded.py`: passes, `sync` → `secret/shared/database` recognized as already-seeded.
- [x] `bats tests/scripts/vault.bats tests/scripts/vault-required-vars-control.bats tests/scripts/vault-empty-guard.bats tests/scripts/vault-covers-compose-control.bats tests/scripts/vault-revoke-order-control.bats`: 129/129 pass, including 10 new/rewritten tests specific to this change (sudo absence pinned, token bound checked against `max_lease_ttl`, the AppRole-login-not-token-renew shape, `cmd_setup_sync_approle` structural checks).
- [x] YAML parses (`yaml.safe_load`) and every `run:` block passes `bash -n`, extracted and checked individually.
- [x] Positive control shown above, captured for real, both failure and success paths, against the live OpenBao.
- [x] `policy-sync.hcl`: untouched, `grep -c sudo` still 0 — pinned by a new bats test so a future edit can't quietly add it.
- [ ] Not deployed. No token minted. No OpenBao config touched. PR only, per instruction — the one-time `setup-sync-approle` step and the merge itself are yours to sequence.

### DIFF
--- a/.github/workflows/vault-sync-to-sops.yml
+++ b/.github/workflows/vault-sync-to-sops.yml
@@ -78,16 +78,31 @@ jobs:
           ssh -i ~/.ssh/vps_key -o StrictHostKeyChecking=no -o ConnectTimeout=10 \
             deploy@${{ steps.get-ip.outputs.tailscale_ip }} 'echo "SSH connection successful"'
 
-      - name: Read sync token from SOPS
-        id: sync-token
+      # h#711: replaces the old "Read sync token from SOPS" step.
+      # VAULT_SYNC_TOKEN was a periodic token that had to be RENEWED weekly —
+      # and renewal silently capped at OpenBao's config.hcl max_lease_ttl=24h
+      # instead of the token's own -period=768h, because periodic-token
+      # renewal only exceeds the system ceiling when the renewing identity
+      # holds `sudo` on auth/token/renew-self, which policy-sync never
+      # granted. The token was dead most weeks; see #711.
+      #
+      # VAULT_SYNC_ROLE_ID/VAULT_SYNC_SECRET_ID replace it. They are AppRole
+      # credentials, not a token — this step only reads them; the next step
+      # logs in fresh every run instead of renewing anything.
+      - name: Read sync AppRole credentials from SOPS
+        id: sync-creds
         run: |
-          VAULT_SYNC_TOKEN=$(sops -d infra/secrets/prod.enc.env | grep '^VAULT_SYNC_TOKEN=' | cut -d= -f2-)
-          if [ -z "$VAULT_SYNC_TOKEN" ]; then
-            echo "::error::VAULT_SYNC_TOKEN not found in SOPS. Run 'vault.sh setup-sync-token' first."
+          PLAIN=$(sops -d infra/secrets/prod.enc.env)
+          ROLE_ID=$(printf '%s\n' "$PLAIN" | grep '^VAULT_SYNC_ROLE_ID=' | cut -d= -f2-)
+          SECRET_ID=$(printf '%s\n' "$PLAIN" | grep '^VAULT_SYNC_SECRET_ID=' | cut -d= -f2-)
+          if [ -z "$ROLE_ID" ] || [ -z "$SECRET_ID" ]; then
+            echo "::error::VAULT_SYNC_ROLE_ID/VAULT_SYNC_SECRET_ID not found in SOPS. Run 'vault.sh setup-sync-approle' first (h#711) — this is expected to fail on the first scheduled run after that migration lands, until the one-time credential mint has happened and been pushed."
             exit 1
           fi
-          echo "::add-mask::$VAULT_SYNC_TOKEN"
-          echo "token=$VAULT_SYNC_TOKEN" >> $GITHUB_OUTPUT
+          echo "::add-mask::$ROLE_ID"
+          echo "::add-mask::$SECRET_ID"
+          echo "role_id=$ROLE_ID" >> $GITHUB_OUTPUT
+          echo "secret_id=$SECRET_ID" >> $GITHUB_OUTPUT
 
       - name: Check script drift
         run: |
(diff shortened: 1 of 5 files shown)


jonhill90:
Verified and ready. **Not merging yet, for a sequencing reason unrelated to the change — see the end.**

Every condition met, and I checked each rather than reading the checklist:

- **No `sudo` introduced.** The only occurrences are explanatory comments about why the periodic-token approach would have needed it and this does not. `policy-sync.hcl` untouched, `grep -c sudo` still 0 — and **pinned by a new bats test**, so a future edit cannot quietly add the estate's first non-break-glass elevated grant. Pinning it is better than my instruction, which only said not to add it.
- **The token bound is in the code, not just the PR body.** `token_ttl=1h`/`token_max_ttl=4h`, comfortably inside `config.hcl`'s `max_lease_ttl=24h`, with the reasoning next to it. The PR body is not what the next person reads.
- **Positive control captured live**, both failure and success paths, read-only against the running OpenBao.
- **The one-time re-mint is stated, and — better — the workflow's own error message says it.** *"this is expected to fail on the first scheduled run after that migration lands, until the one-time credential mint has happened."* Whoever is on call that week reads that instead of diagnosing a regression. That is the difference between documenting a caveat and building it into the thing that will surface it.

**Why I am holding.**

A lane exposed Hill90's entire production SOPS store in plaintext into its own transcript earlier — a `sops -d` without the right `--input-type` producing empty output, then a diff against that empty output printing all 77 lines. Contained: the VPS scratch directory with two more plaintext copies is shredded and verified gone, the committed store is untouched, nothing pushed. But the exposed set includes `VAULT_DB_/AUTH_/INFRA_/OBSERVABILITY_SECRET_ID` — the four live AppRole secret-ids this mechanism depends on.

That does not make this change wrong. It makes it **one decision instead of two**: this PR is inert until a human runs `vault.sh setup-sync-approle`, and that mint is exactly the kind of step that should be folded into whatever credential rotation Jon decides on, rather than done twice a few hours apart.

So this waits on Jon's rotation call — not on anything in the diff. If he decides no rotation is warranted, merge it as-is and do the mint on its own.

PR #735, opened 2026-08-05
fix(checks): close two vacuous-pass shapes found in a deliberate sweep

jonhill90:
Follow-up to #724/#725: rather than wait for the next red deploy to find another instance of "a check that runs against production and can pass for the wrong reason" by accident, swept `scripts/checks/` (33 files) and the verification steps in `.github/workflows/` (9 files) deliberately, hunting four shapes: a discarded curl/network exit status read as a specific verdict; a grep/comparison whose empty result is treated as agreement (the 0==0 trap); a check that derives its "expected" value the same way as the thing under test; and a check with no readiness gate right after something is recreated.

## Ranked findings — 11 real, ~40 files triaged

**Fixed here (both positive-controlled in both directions against real services):**

1. **`scripts/checks/check_destructive_commands.sh`** — the 0==0 trap. `for f in scripts/*.sh .github/workflows/*.yml` with neither glob matching leaves the pattern unexpanded, every iteration `continue`s, `FAIL` never leaves 0 — "No destructive volume commands found" having scanned nothing. Reproduced directly (empty dir → exit 0, that message). Guards a security-relevant invariant CLAUDE.md calls load-bearing. Latent, not live, today — its only caller always runs from repo root — one working-directory edit away.
2. **`.github/workflows/scheduled-checks-watchdog.yml`**'s Alertmanager notify-log step — could not fail regardless of what it found (`grep ... || echo ...` always exits 0). This is the watchdog's own last-mile confirmation that an alert reached Alertmanager — the layer meant to page a human when every OTHER scheduled check goes red. Verified against a real `prom/alertmanager:v0.28.1` container (the pinned prod version): captured a genuine delivery-failure log line, confirmed the OLD step logic exits 0 against it, confirmed the NEW logic exits 1. Also discovered while building the fixture and worth recording: Alertmanager logs **nothing** matching `notify|smtp|email` at info verbosity for a *successful* delivery — confirmed against the same container with a working receiver — so "no matching lines" correctly stays a soft pass.

Both carry `tests/scripts/*.bats` control tests built from real captured fixtures (not synthetic guesses), following this repo's existing convention. Confirmed failing-test-first: stashing each fix re-fails 3/7 and 4/5 of the new assertions respectively.

**Filed as issues, ranked by consequence, not fixed here:**

3. [#726](https://github.com/jonhill90/Hill90/issues/726) `check_root_revoke_fails_closed.py` never invoked anywhere — a root-token safety net wired to nothing. Highest-consequence finding in the sweep; not fixed here because wiring a check against the real root-minting workflows deserves its own review, not a bundle.
4. [#727](https://github.com/jonhill90/Hill90/issues/727) `check_retired_roles_ungranted.py --live` discards SSH failures into a vacuous pass, and also has no caller anywhere.
5. [#728](https://github.com/jonhill90/Hill90/issues/728) `check_secrets_schema.py` always exits 0 in CI regardless of real findings — live today, but the fix is a policy decision (flip `SECRETS_SCHEMA_STRICT`), not a bug fix, so filed rather than decided unilaterally.
6. [#729](https://github.com/jonhill90/Hill90/issues/729) `check_role_mappings_repointed.py` silently drops a site if its role-list regex extracts empty.
7. [#730](https://github.com/jonhill90/Hill90/issues/730) `check_declared_paths_are_seeded.py` has no guard for `declared_paths()` returning empty (unlike its sibling `check_oidc_clients_reconciled.py`).
8. [#731](https://github.com/jonhill90/Hill90/issues/731) `vault-sops-name-parity.sh` can report a real vault path as ABSENT on a transient docker-exec failure — false-RED, not false-green, lower priority.
9. [#732](https://github.com/jonhill90/Hill90/issues/732) `ci.yml`'s "Assert deploy safety invariants" step has a dormant 0==0 trap (functions correctly today, 22 real matches).
10. [#733](https://github.com/jonhill90/Hill90/issues/733) `vault-reinitialize.yml`'s post-restart check uses fixed sleeps instead of polling — fails in the safe direction, lowest priority of the readiness-gate findings.
11. [#734](https://github.com/jonhill90/Hill90/issues/734) `check_md_links.py` has a cosmetic-blast-radius version of finding #1.

**Checked and correctly built, named so nobody re-checks them:** `check_alert_series.py`, `check_alert_counts_documented.py`, `check_checks_are_wired.py`, `check_argv_secrets.sh`, `check_deploy_drift.sh`, `check_tenant_baseline_agrees.py`, `check_vault_revoke_order.py`, `check_hill90_ui_secret_agreement.sh`, `check_vault_covers_compose.py`, `check_oidc_clients_reconciled.py`, `check_realm_tenant_clients.py`, `check_env_surface.py`, `check_edge_middlewares.py`, `vault-oidc-enabled-test.sh`, `minio-readiness-test.sh`, `minio-policy-names-test.sh`, `platform-baseline-test.sh`, `vault-approle-login-test.sh`, `vault-approle-real-read-test.sh`, `vault-oidc-login-test.sh`, `realm-tenant-serves-test.sh`, `tenant-db-isolation-test.sh`, `tenant-login-local-test.sh`, `audit-hill90-ui-client.yml`, `deploy-drift.yml`, `reusable-deploy-service.yml`'s other steps, `tenant-baseline-agreement.yml`, `vault-init.yml`, `vault-regain-root.yml` — each already carries an explicit vacuous-pass guard, fails closed, or is properly gated by an earlier readiness step.

## Test plan

- [x] `check_destructive_commands.sh`: `tests/scripts/destructive-commands-control.bats`, 7/7 pass; 3/7 fail against pre-fix code (confirmed via stash).
- [x] `scheduled-checks-watchdog.yml`: `tests/scripts/watchdog-alertmanager-notify-check.bats`, 5/5 pass, built from real fixtures captured off a live `prom/alertmanager:v0.28.1` container; 4/5 fail against pre-fix code (confirmed via stash).
- [x] Full `bats tests/scripts/` suite: 542 tests, all green.
- [x] `shellcheck` clean on `check_destructive_commands.sh`.
- [x] YAML validated (`yaml.safe_load`).

No infra/secrets/sops/credential work. No deploy run from a workstation.

### DIFF
--- a/.github/workflows/scheduled-checks-watchdog.yml
+++ b/.github/workflows/scheduled-checks-watchdog.yml
@@ -180,8 +180,48 @@ jobs:
       # This step names exactly what it can see: whether Alertmanager's own
       # notification pipeline logged an attempted/successful send for THIS
       # alert. It cannot see an inbox, and does not claim to.
+      # THIS STEP USED TO BE UNABLE TO FAIL, REGARDLESS OF WHAT IT FOUND.
+      # `grep ... || echo ...` means the step's exit code is whichever of
+      # grep or echo ran — grep exits 0 the moment it matches ANY line, so a
+      # genuine "Notify attempt failed, will retry later" WARN line (the real
+      # shape Alertmanager logs on delivery failure, confirmed against a real
+      # prom/alertmanager:v0.28.1 container — the version this estate pins in
+      # docker-compose.observability.yml) still made this step GREEN. Verified
+      # directly: captured that exact log line from a real container whose
+      # webhook receiver was unreachable, ran the old `grep ... || echo` line
+      # against it, exit 0. This step is the last-mile confirmation in the
+      # watchdog that pages a human when every OTHER scheduled check goes red
+      # — a check that cannot fail here is decorative at the one layer this
+      # workflow exists to be the backstop for.
+      #
+      # WHAT SUCCESS LOOKS LIKE IN THESE LOGS: nothing. Alertmanager does not
+      # log a line for a notification it delivered successfully at its
+      # default (info) verbosity — confirmed against the same real container:
+      # a webhook receiver that accepted every POST produced zero matching
+      # log lines, ever, in repeated trials. So "no matching lines" cannot be
+      # promoted to a hard failure without false-paging on every clean run;
+      # it stays informational, exactly as the comment above this step
+      # already says ("It cannot see an inbox, and does not claim to").
+      #
+      # WHAT CHANGED: a matched line is now inspected, not just printed. If
+      # it contains a failure indicator (Alertmanager's own retry-failure
+      # vocabulary, confirmed against the real log line above: "failed"), the
+      # step now fails loud with ::error::. A benign match with no failure
+      # indicator, or no match at all, stays a soft pass — that boundary is
+      # unchanged from before.
       - name: Check Alertmanager's own notify log for this alert
         run: |
           ssh -i ~/.ssh/vps_key -o StrictHostKeyChecking=no -o ConnectTimeout=15 \
-            deploy@${{ steps.get-ip.outputs.tailscale_ip }} \
-            'docker logs --since=2m alertmanager 2>&1 | grep -i "notify\|smtp\|email" || echo "(no matching log lines in the last 2 minutes)"'
+            deploy@${{ steps.get-ip.outputs.tailscale_ip }} '
+              LOG_LINES="$(docker logs --since=2m alertmanager 2>&1 | grep -iE "notify|smtp|email" || true)"
+              if [ -z "$LOG_LINES" ]; then
+                echo "(no matching log lines in the last 2 minutes — cannot confirm a send was attempted; Alertmanager does not log successful notify attempts at this verbosity)"
+                exit 0
+              fi
+              echo "$LOG_LINES"
+              if echo "$LOG_LINES" | grep -qiE "failed|error|refused|timeout|unable|denied|unauthorized"; then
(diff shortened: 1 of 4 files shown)


jonhill90:
Verified the two fixes against the files, and merging.

`check_destructive_commands.sh` now counts what it scanned and exits **2** with "CANNOT DETERMINE" when the globs match nothing, rather than falling through to "No destructive volume commands found" having read zero files. Reporting the scanned count in the success message is the better half of that fix — it makes a vacuous pass visible even to someone not looking for one. And exit 2 is the right code: this is neither a pass nor a violation, it is the instrument failing to see, and collapsing that into either one is how the original class of bug survives. Worth noting it guards an invariant CLAUDE.md calls load-bearing, and was one working-directory change away from being live rather than latent.

The watchdog's notify step now distinguishes three states instead of one. `grep ... || echo ...` could not fail regardless of what it read — on the last-mile confirmation that an alert reached Alertmanager, the layer whose whole job is paging a human when every other scheduled check goes red. Keeping "no matching lines" as a soft pass is correct and, importantly, was **verified rather than assumed**: Alertmanager logs nothing at that verbosity for a successful delivery, confirmed against a real `prom/alertmanager:v0.28.1` with a working receiver. Building both fixtures off the pinned production version rather than writing synthetic log lines is what makes the control mean something.

The "checked and correctly built" list of 29 is worth as much as the fixes. A sweep that reports only its findings leaves the next sweep re-deriving every negative result; naming what was checked and found sound is what stops that.

Nine remaining findings filed individually rather than bundled, ranked by consequence. h#726 is already dispatched.

PR #739, opened 2026-08-05
fix(checks): stop check_retired_roles_ungranted.py's --live mode from turning an unreachable host into a PASS

jonhill90:
Closes #727.

## The bug

`live_checks()` reached the VPS via `ssh(cmd) or "[]"` — when the ssh call itself failed, `.stdout` was empty, `"" or "[]"` became a valid empty JSON list, and the `except` around `json.loads` never fired. An unreachable host silently reported **PASS**, having measured nothing. The same shape recurred for the MinIO policy read.

The instruction for this fix named the exact hazard: an unreachable host must not read as "no retired role is granted" — that's the answer that agrees with what everyone hopes, which is precisely the wrong direction for a security check to fail in.

## The fix

Both `ssh()` calls now capture the full `CompletedProcess`, check `returncode` explicitly, and raise a new `LiveCheckUnreachable` — distinct from both PASS and FAIL — on failure. `main()` catches it and returns **exit 2** (CANNOT DETERMINE), matching the sibling convention already used by `check_deploy_drift.sh` and `check_tenant_baseline_agrees.py`. Three genuinely different exit codes now: 0 (clean), 1 (real violation), 2 (could not measure).

## Positive-controlled both directions, against something real

Not mocked — a throwaway `linuxserver/openssh-server` container, with the script's literal hardcoded `ssh vps` alias temporarily pointed at it via a `Host vps` block placed ahead of the real one in `~/.ssh/config` (restored afterward — diffed clean against a backup taken first; the container and its key material lived only in `/tmp` for the test's duration).

| Scenario | Result |
|---|---|
| Clean live estate | **PASS**, exit 0 |
| A retired role genuinely present, live, in both Keycloak and MinIO | **FAIL**, exit 1, names both |
| Remote command chain fails (stand-in `docker exec` returns real non-zero) | **CANNOT DETERMINE**, exit 2 |
| Host genuinely unreachable (container stopped, real "Connection refused") | **CANNOT DETERMINE**, exit 2 |

**The decisive comparison** — same real stopped container, only the code differed:
```
$ # pre-fix code, real unreachable host
PASS — no retired role is granted anything
OLD CODE exit=0

$ # fixed code, same real unreachable host
CANNOT DETERMINE — could not read live realm roles — ssh/remote command exited 255: ssh: connect to host localhost port 12222: Connection refused
exit=2
```

`tests/scripts/retired-roles-ungranted-control.bats` gained a permanent, CI-safe regression form of the same four scenarios via a PATH-shadowing `ssh` stub (no Docker/network dependency at test time) — the real-container run is what established the fix is correct; the stub tests are what keep it that way. Failing-test-first confirmed: 3 of the 5 new assertions fail against the pre-fix script.

## Also wired: static mode

`python3 scripts/checks/check_retired_roles_ungranted.py` (no `--live`) is explicitly documented in the script's own docstring as the CI-safe half ("static, CI") — needs no secrets, now runs in `ci.yml`. It cannot see live holders, so it does not replace `--live`.

## Stopped at a real boundary: `--live` itself is not wired anywhere yet

Filed as **#738**. `--live` hardcodes `ssh vps <cmd>` — an SSH config alias literal that nothing in this repo's CI establishes; every workflow reaches the VPS via `ssh -i ~/.ssh/vps_key ... deploy@$TAILSCALE_IP` directly, no alias. Wiring it needs a real decision (add an alias-setup step that duplicates what every other workflow does differently, or refactor the script to take an explicit SSH target) and validating either path needs a live run against production Keycloak/MinIO with decrypted secrets — explicitly out of scope for this task.

## Test plan

- [x] Real-container positive control, both directions, shown above.
- [x] `tests/scripts/retired-roles-ungranted-control.bats`: 12/12 pass; 3/12 fail against pre-fix code (stashed and re-run).
- [x] `check_checks_are_wired.py` now reports `check_retired_roles_ungranted.py` as `bats,workflow`.
- [x] Full `bats tests/scripts/` suite: 535 tests, all green.
- [x] YAML validated.

Not infra/secrets/sops/credential work. No token/credential minted, rotated, or written. No deploy run from a workstation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,29 @@ jobs:
       - name: Assert every referenced OIDC client is created on the deploy path
         run: python3 scripts/checks/check_oidc_clients_reconciled.py
 
+      # h#727: static mode (no --live) is explicitly documented in the
+      # script's own module docstring as "static, CI" — this wires that
+      # documented, secrets-free half in. It cannot see holders (needs the
+      # live realm), so it cannot replace --live; it catches the config-file
+      # half of the same property on every PR: a retired role must not be
+      # (re)created in keycloak.sh, or named in MinIO's policy list,
+      # Grafana's role_attribute_path, or OpenBao's bound_claims.
+      #
+      # --live is NOT wired here, or anywhere yet, and that is a deliberate
+      # stop rather than an oversight: it hardcodes `ssh vps <cmd>` — an SSH
+      # config alias literal — and no workflow in this repo defines a "vps"
+      # host alias; every other workflow reaches the VPS via
+      # `ssh -i ~/.ssh/vps_key ... deploy@$TAILSCALE_IP` directly, with no
+      # alias at all. Wiring --live needs either a new alias-setup step here
+      # (a real design decision, not obviously the right one — it duplicates
+      # what every other workflow already does differently) or refactoring
+      # the script to accept an explicit SSH target instead of a hardcoded
+      # alias, and validating either path end-to-end needs a real run against
+      # production Keycloak/MinIO with decrypted secrets, out of scope for
+      # this change. See h#738 for what's left.
+      - name: Assert no retired role is granted anything (static)
+        run: python3 scripts/checks/check_retired_roles_ungranted.py
+
       - name: Assert no path revokes root before OIDC is enabled
         run: python3 scripts/checks/check_vault_revoke_order.py
 
--- a/scripts/checks/check_retired_roles_ungranted.py
+++ b/scripts/checks/check_retired_roles_ungranted.py
@@ -4,6 +4,13 @@
     python3 scripts/checks/check_retired_roles_ungranted.py           # static, CI
     python3 scripts/checks/check_retired_roles_ungranted.py --live    # + live estate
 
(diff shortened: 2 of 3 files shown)


jonhill90:
Verified against the file and merging. `ssh(cmd) or "[]"` is the entire bug in one expression: a failed SSH left stdout empty, `"" or "[]"` became a syntactically valid empty JSON list, and the `json.loads` guard downstream never had anything to catch. The check then answered "are any retired roles granted" with "no", having looked at nothing.

The direction is what makes it serious. "No retired role is granted" is the answer everyone hopes for, so this failed toward agreement, on a security property — the dangerous null CONTRIBUTING already names.

Three exit codes is the right resolution (0 clean, 1 violation, 2 could-not-measure), and reusing `check_deploy_drift.sh`'s existing convention rather than inventing a fourth spelling keeps the estate legible. Proving CANNOT DETERMINE against a genuinely stopped container producing a real "Connection refused", not a mocked non-zero, is the difference between testing the code path and testing the condition. Backing up `~/.ssh/config` and diffing it clean afterwards was the right instinct for a test that reaches into your own machine's config. Wired into `ci.yml`, so it can no longer be a check that exists without ever running.

PR #741, opened 2026-08-05
fix(checks): refuse a vacuous pass when declared_paths() finds no arms (h#730)

jonhill90:
Closes #730 — steals the guard from its sibling `check_oidc_clients_reconciled.py` rather than inventing a new shape, per the instruction.

## The bug

`declared_paths()` parses the case arms of `vault_paths_for_service()` with a regex requiring the literal shape `arm)  echo "..."`. If that function is ever refactored away from `echo` — a case arm rewritten to `printf`, for instance — the outer "could not find `vault_paths_for_service()`" guard still matches (the function body itself is still found), but the per-arm regex silently matches zero arms. The check printed "PASS — every declared path is seeded" having examined none of them.

## The fix

`check_oidc_clients_reconciled.py` already has the identical guard for the identical shape:

```python
if not refs:
    print("FAIL — no client references found at all. The patterns no longer match "
          "anything, which is a broken check, not a clean estate.", file=sys.stderr)
    return 1
```

Copied it — same wording, same exit behavior — rather than writing something new:

```python
if not declared:
    print("FAIL — no declared paths found at all. The pattern no longer matches "
          "anything in vault_paths_for_service(), which is a broken check, not a "
          "clean estate.", file=sys.stderr)
    return 1
```

## Positive-controlled directly

Ran the **old** script against a real copy of `scripts/_common.sh` with every arm's `echo "..."` mechanically rewritten to `printf '%s' "..."` — the exact drift the fix targets:

```
$ python3 check_declared_paths_are_seeded.py   # old code, mutated real source
Declared vault paths must be seeded
===================================

  note: seeded but declared by no service: secret/auth/config, secret/infra/traefik, secret/infra/vps, secret/observability/grafana, secret/shared/database

PASS — every declared path is seeded, and every seeded key is guarded
exit=0
```

Zero of the six real declared paths were examined — visible only as "seeded but declared by no service" *orphans*, never as a single `ok`/`MISS` line, and still a clean PASS. The **new** code, same mutation:

```
FAIL — no declared paths found at all. The pattern no longer matches
anything in vault_paths_for_service(), which is a broken check, not a
clean estate.
exit=1
```

The real, unmutated `scripts/_common.sh` is unaffected — still `PASS`, all six paths examined as before.

## Test plan

- [x] `tests/scripts/declared-paths-seeded-control.bats` gained a permanent regression test for this exact mutation — 8/8 pass; the new assertion fails against the pre-fix script (stashed and re-run).
- [x] Full pytest suite: 75/75 pass.
- [x] Full `bats tests/scripts/`: 543/543 pass.

Not infra/secrets/sops/credential work. No deploy run from a workstation.

### DIFF
--- a/scripts/checks/check_declared_paths_are_seeded.py
+++ b/scripts/checks/check_declared_paths_are_seeded.py
@@ -89,6 +89,19 @@ def seeded_keys(path: str) -> set[str]:
 
 def main() -> int:
     declared = declared_paths()
+
+    # h#730: declared_paths() returning empty is indistinguishable from every
+    # case arm in vault_paths_for_service() having rotted out of the regex,
+    # and would pass vacuously — "every declared path is seeded" having
+    # checked zero of them. Refuse instead, matching the sibling check this
+    # one is modelled on: check_oidc_clients_reconciled.py's identical guard
+    # on `refs` a few lines into its own main().
+    if not declared:
+        print("FAIL — no declared paths found at all. The pattern no longer matches "
+              "anything in vault_paths_for_service(), which is a broken check, not a "
+              "clean estate.", file=sys.stderr)
+        return 1
+
     seeded = seeded_paths()
     required = required_keys()
 
--- a/tests/scripts/declared-paths-seeded-control.bats
+++ b/tests/scripts/declared-paths-seeded-control.bats
@@ -110,6 +110,39 @@ PY
   [[ "$output" == *"empty string"* ]]
 }
 
+# h#730: unlike its sibling check_oidc_clients_reconciled.py (which refuses a
+# zero-reference result as indistinguishable from every pattern having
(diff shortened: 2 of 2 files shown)


jonhill90:
Merging. Reusing `check_oidc_clients_reconciled.py`'s existing guard rather than writing a third spelling of the same idea is the right instinct — the estate now has one recognisable shape for "I examined nothing", and a reader who has seen it once recognises it everywhere.

The failure mode you describe is the subtle half of this family and worth stating plainly: the outer guard still passed, because `vault_paths_for_service()` was still found. Only the inner per-arm regex went to zero. A check with a guard on the wrong granularity looks guarded — that is exactly why it survives review — and the giveaway is always the same, a count that nobody asserts is non-zero.

PR #742, opened 2026-08-05
fix(checks): close the remaining 0==0-family findings from h#735's sweep (h#729, h#732, h#734)

jonhill90:
Closes #729, #732, #734.

Same family as #726/#727/#730/#735's `check_destructive_commands.sh` fix: "examined nothing" silently reading as "examined everything and found nothing." Bundled into one PR since all three are genuinely that shape — reused the two conventions this repo already established rather than inventing per-site variants:

- **Shape A** — zero inputs matched at all → **CANNOT DETERMINE**, distinct from a clean pass (precedent: `check_destructive_commands.sh`).
- **Shape B** — input was read, a match was found, but parsing *within* that match yielded nothing → a **per-finding FAIL**, not a separate exit code (precedent: `check_declared_paths_are_seeded.py`, `check_oidc_clients_reconciled.py`).

## #734 (shape A) — `check_md_links.py`

If every `SCAN_ROOTS` entry stops existing, the script scanned zero files and still printed "no missing local links found." Now counts files scanned; zero is `CANNOT DETERMINE` (exit 2) — the exact fix the issue asked for, matching `check_destructive_commands.sh`.

Positive-controlled: an isolated tree with none of the four scan roots → refuses (exit 2). Same tree + one real broken link → still fails (exit 1). A genuinely clean non-empty scan → still passes, now reports its scan count.

## #732 (shape A — **plus a more severe defect found while fixing it**) — `ci.yml`

The filed defect: the `-p`-invariant grep matches zero lines if `deploy.sh`'s invocation style is ever renamed, and the pipeline finds nothing to complain about.

**Found while fixing it, more severe:** GitHub Actions runs `run:` blocks under `bash -eo pipefail`, and bash's `set -e` explicitly *exempts* a `!`-negated command from triggering abort — even when the negated result is failure. `! (pipeline)` never aborted this step on a **real** violation; it fell through to the final `echo` and always exited 0. This wasn't dormant — the step could not fail today for any real violation of either invariant, full stop. Verified under the exact bash flags GitHub Actions uses, not by reasoning about semantics:

```
$ # old step, real deploy.sh + a genuine docker-compose call with no -p
Checking: all docker compose calls use -p...
  docker compose down --remove-orphans
All safety checks passed.
exit=0
```

Rewrote both checks with explicit `if`/`exit` (fixes the `!`-under-`set -e` defect for both, independent of the 0==0 trap) and added the zero-matches `CANNOT DETERMINE` guard to the `-p` check. Positive-controlled all four states against real `deploy.sh` mutations: clean passes; a real `--remove-orphans` violation fails; a real missing-`-p` violation fails; zero `docker compose` invocations refuses. The old step's inability to fail on real violations is kept as a permanent bats regression-guard test.

## #729 (shape B — decided deliberately, as asked) — `check_role_mappings_repointed.py`

If the outer regex matches a mapping expression but the inner one extracts zero role names (e.g. single quotes → double quotes), the loop silently `continue`d — not counted, not printed, not caught by the existing "gone blind" guard (which only fires when the *outer* regex matches nothing).

**The decision, and why:** per-site **FAIL**, not a whole-run `CANNOT DETERMINE`.
1. The input *was* read and a real match *was* found — narrower than "nothing examined," closer to `check_declared_paths_are_seeded.py`'s shape (which uses FAIL/exit 1) than to `check_retired_roles_ungranted.py --live` or `check_destructive_commands.sh`, where nothing was read at all.
2. This file already has a per-site convention at exactly this granularity — the existing "gone blind" guard is per-site, not whole-run.
3. Per-site reporting preserves more information: if one of two sites goes blind while the other is genuinely clean, naming exactly which site and why beats declaring the whole run unmeasurable — and the whole-run "examined nothing" case is still correctly caught as a consequence, since every site failing to extract now contributes its own FAIL entry.

Positive-controlled against a real mutation of the real `docker-compose.observability.yml` (single→double quotes in Grafana's `role_attribute_path`, the exact drift class the issue names):

```
$ # OLD code, real mutation
PASS — 1 mapping(s) checked, none names a legacy realm role
exit=0   # Grafana silently dropped entirely, zero trace

$ # NEW code, same real mutation
FAIL — 1 problem(s):
  ...Grafana role_attribute_path matched a mapping expression but no role
  names could be extracted from it...
exit=1
```

Also confirmed a genuinely legacy-named *other* site is still independently reported even when one site is unparseable, and that real legacy-role detection is unaffected.

## Test plan

- [x] All three positive-controlled failing-test-first: stashed each fix, reran its new tests against pre-fix code — 4/7, 4/7, 2/8 fail as expected; all pass against the fix.
- [x] Full pytest suite: 80/80 pass.
- [x] Full `bats tests/scripts/`: 556/556 pass.
- [x] YAML validated.

## What's still open from h#735's ranked list

#731 (false-RED on a transient docker-exec failure) and #733 (fixed sleeps instead of polling) — deliberately left, per the sweep's own ranking and this task's scope.

Not infra/secrets/sops/credential work. No deploy run from a workstation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,13 +71,44 @@ jobs:
       - name: Lint scripts
         run: shellcheck --severity=error scripts/*.sh
 
+      # h#732 found the dormant 0==0 trap on the second check below (if
+      # deploy.sh's invocation style is ever renamed, the grep for "docker
+      # compose" matches zero lines and the check finds nothing to complain
+      # about — an unchecked invariant, not an upheld one). Fixing it
+      # surfaced something more severe in the SAME lines: GitHub Actions
+      # runs `run:` blocks under `bash -eo pipefail`, and bash's `set -e`
+      # explicitly EXEMPTS a `!`-negated command from triggering abort, even
+      # when the negated result is failure. `! (pipeline)` never aborted
+      # this step on a real violation — it just fell through to the next
+      # line, and the step always ended on the final `echo`'s exit 0.
+      # Verified directly, not by reasoning about bash semantics alone:
+      # injected a real `docker compose ... ` call with no -p flag into a
+      # copy of deploy.sh and ran this exact step, byte for byte, under
+      # `bash -eo pipefail` — it printed "All safety checks passed." and
+      # exited 0. Both checks are rewritten with an explicit if/exit so a
+      # real violation actually fails the step, which the `!`-negated form
+      # could not do regardless of the 0==0 trap.
       - name: Assert deploy safety invariants
         run: |
           echo "Checking: no --remove-orphans..."
-          ! (grep -- "--remove-orphans" scripts/deploy.sh | grep -v "^#" | grep -v "^[[:space:]]*#")
+          if grep -- "--remove-orphans" scripts/deploy.sh | grep -v "^#" | grep -v "^[[:space:]]*#" | grep -q .; then
+            echo "::error::scripts/deploy.sh contains --remove-orphans"
+            grep -n -- "--remove-orphans" scripts/deploy.sh | grep -v "^#"
+            exit 1
+          fi
 
           echo "Checking: all docker compose calls use -p..."
-          ! (grep "docker compose" scripts/deploy.sh | grep -v "^#" | grep -v "^[[:space:]]*#" | grep -v -- "-p ")
+          DC_LINES=$(grep "docker compose" scripts/deploy.sh | grep -v "^#" | grep -v "^[[:space:]]*#" || true)
+          if [ -z "$DC_LINES" ]; then
+            echo "::error::CANNOT DETERMINE — zero 'docker compose' invocations found in scripts/deploy.sh. Either the invocation pattern changed (renamed, wrapped, hyphenated) and this check is no longer looking at anything real, or deploy.sh itself changed shape. Fix this check before trusting a green run here."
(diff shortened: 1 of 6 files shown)


jonhill90:
Merging. I verified the bash claim independently before accepting it, because it is a strong statement about a load-bearing invariant: under `bash -eo pipefail`, a `!`-negated pipeline whose negation is false does **not** abort, and the script runs on to exit 0. Reproduced directly rather than reasoned about.

That settles the severity. CLAUDE.md invariant 2 says `--remove-orphans` is banned globally "and CI enforces it." It did not. This was not the dormant 0==0 trap the issue described — it was a guard that could not fire for any real violation of either invariant. Finding the worse defect while fixing the milder one and reporting it as its own thing, rather than folding it quietly into the diff, is the right call.

The Shape A / Shape B distinction is the useful generalisation out of this whole sweep, and worth keeping: **zero inputs matched at all** is CANNOT DETERMINE, because the instrument saw nothing; **input read, match found, parsing within it yielded nothing** is a per-finding FAIL, because the instrument worked and the data is wrong. Those get conflated constantly, and naming both with the precedent already in this repo means the next check picks the right one without re-deriving it.

The invariant text and the separate question of whether anything actually reached the tree during the window are in flight separately — the guard being off and something getting through are different facts, and I do not want the second assumed from the first.

PR #743, opened 2026-08-05
docs(claude): correct invariant 2 — CI never actually enforced --remove-orphans

jonhill90:
Follow-up to #742: fixing the check made "CI enforces it" true again, which means it wasn't true before — and the invariant text should record what actually happened, not quietly become accurate.

## The window, established from git history, not assumed

The `!`-negated pipeline was broken from the check's very **first** commit — `3685dcb8` (2026-02-22, in the then-separate `ci-deploy-scripts.yml`), folded into `ci.yml` the next day by `220d28f0`. CLAUDE.md's own claim was written on `a5685cf3` (2026-07-29) — **over five months after the guard was already non-functional.** The claim was never true at any point it existed.

The scope claim was also wrong independent of the exit-code bug: the check has only ever grepped `scripts/deploy.sh`, never `scripts/local.sh` or any workflow file — so "banned globally, in every script and workflow" overstated what was ever even *attempted*, let alone enforced.

## Did anything actually get through? Checked, not assumed.

`--remove-orphans` has not appeared in `scripts/deploy.sh` — the only file the check has ever examined — since `e72f93e3` removed it on **2026-02-14, eight days before the check was even introduced.** The guard was off for the check's entire existence; nothing exploited it in the one file it covers, because nothing tried.

`scripts/local.sh` does use `--remove-orphans` in six places (added 2026-07-26) — but exclusively against Docker-Desktop-only compose projects prefixed `hill90-local-`, the same local-only scope `check_destructive_commands.sh` already documents as an exemption for a different invariant on this same file. Not a violation of what this invariant actually exists to prevent (cross-stack teardown on shared/production infrastructure) — but real, current usage the old "every script" wording didn't account for.

## The fix

Corrected in the same self-correction style invariant 1 already uses in this file: keep the rule, append what was actually true with dates and commit hashes, distinguish "the guard was off" from "something got through" rather than letting either overstate the other.

Docs only. Not infra/secrets/sops/credential work.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -90,7 +90,9 @@ docs/                   runbooks, reference, architecture, decisions
    whole time, independent of the exit-code bug.** The check has only ever
    grepped `scripts/deploy.sh` — never `scripts/local.sh`, never any
    `.github/workflows/*.yml` file. `scripts/local.sh` uses `--remove-orphans`
-   in six places (added 2026-07-26), which is why this line no longer claims
+   in **twelve** places (added 2026-07-26) — six services across two teardown
+   paths, `down` (lines 444, 446, 448, 450, 452, 454) and `down -v` (lines
+   495, 496, 497, 498, 500, 501) — which is why this line no longer claims
    "every script": that usage runs exclusively against Docker Desktop
    compose projects prefixed `hill90-local-` (the same local-only scope
    `check_destructive_commands.sh`'s volume-destruction exemption already
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,9 +61,42 @@ docs/                   runbooks, reference, architecture, decisions
    original text above omits entirely. `git log -S` traces `scripts/minio.sh`'s
    addition to #593 — old enough that a prior sweep (#685) should have caught
    it and didn't. Check the workflow's own `on.push.paths`, not this list.
-2. **`--remove-orphans` is banned globally**, in every script and workflow, and
-   CI enforces it. It will delete containers belonging to another stack that
+2. **`--remove-orphans` is banned in `scripts/deploy.sh`**, and CI enforces
+   it there. It will delete containers belonging to another stack that
    shares a Compose project name.
+   **"CI enforces it" was false from the day this line was written, and had
+   been false since before that — `Verified 2026-08-05` against the actual
+   git history, not assumed from the check existing.** The assertion step
+   (`.github/workflows/ci.yml`, "Assert deploy safety invariants") wrote its
+   guard as `! (pipeline)`. GitHub Actions runs `run:` blocks under
+   `bash -eo pipefail`, and bash's `set -e` explicitly exempts a
+   `!`-negated command from triggering abort even when the negated result is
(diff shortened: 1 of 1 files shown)


jonhill90:
The history work is right and I checked the parts I could check independently. `deploy.sh` has zero `--remove-orphans` today, and the six `compose_*` helpers do scope to `hill90-local-*` project names (`EDGE_PROJECT=hill90-local-edge` and siblings, local.sh:26-31), so the "local-only, not cross-stack teardown" reasoning holds.

**One number is wrong, and it is the kind this PR exists to get right.** You say `local.sh` uses `--remove-orphans` "in six places". It is **twelve**:

```
444,446,448,450,452,454   compose_{minio,vault,auth,db,obs,edge} down --remove-orphans
495,496,497,498,500,501   compose_{minio,vault,auth,db,obs,edge} down -v --remove-orphans
```

Six services, two teardown paths each — a plain `down` and a `down -v`. Whoever counted saw one block and stopped. The conclusion does not change, since all twelve are the same local-only helpers, but the count goes into a document being corrected specifically because it carried a number nobody re-counted. Shipping a fresh miscount into that paragraph would be the same failure one turn later.

Fix the number, and while you are there say twelve across two teardown paths rather than a bare twelve — a reader who greps and finds two groups should not have to wonder whether the second was missed again.

Everything else stands, and the strongest part is worth keeping prominent: the claim was **never** true. The negated pipeline was broken from the check's first commit `3685dcb8`, and CLAUDE.md's assertion was written on `a5685cf3` five months later. That is not a guard that decayed, it is a guard that never worked, described as working. And the separation you kept — the guard was off, nothing got through, because nothing tried — is exactly right; neither half should be allowed to imply the other.

jonhill90:
Fixed — confirmed 12, not 6, by `grep -n "remove-orphans" scripts/local.sh` before acting on it: 444/446/448/450/452/454 (`down`) and 495/496/497/498/500/501 (`down -v`), exactly matching your line numbers. Reworded as "six services across two teardown paths" so the two groups of six read as intentional, not a second missed count. Pushed in 80be3dad.

jonhill90:
Count corrected to twelve across two teardown paths — verified in the diff. Merging.

The finding this records is worth stating once more plainly, because it is stronger than the issue that prompted it. This was not a guard that decayed: the `!`-negated pipeline was broken in the check's **first** commit (`3685dcb8`, 2026-02-22), and CLAUDE.md's "CI enforces it" was written on `a5685cf3` five months later. The claim was never true at any point it existed. A reader taking the invariant at face value would have believed an enforcement that had never once run.

And the discipline that keeps it honest in the other direction: nothing got through. `--remove-orphans` left `deploy.sh` on 2026-02-14, eight days before the check existed — I confirmed it is still absent — and the twelve `local.sh` uses are scoped to `hill90-local-*` Docker Desktop projects, which I checked at `local.sh:26-31`. The guard being off and something exploiting it are different facts, and this keeps them separate rather than letting the scarier one be inferred from the true one.

Also correct to flag that the scope claim was independently wrong: "every script and workflow" overstated what was ever attempted, since the check has only ever read `deploy.sh`.

PR #744, opened 2026-08-05
fix(checks): stop vault-sops-name-parity.sh reporting a live path ABSENT on a transient docker-exec failure (h#731)

jonhill90:
Closes #731. Same conflation as #726/#727, opposite direction — a false-RED instead of a false-GREEN — and the task asked for the same three-outcome treatment.

## The bug

`vault_keys()` discarded `docker exec`'s stderr entirely (`2>/dev/null`), so a transient failure and a genuinely absent path both produced empty stdout and were reported identically: `ABSENT ... not present on the vault`.

## What I established before fixing it

Against a real `ghcr.io/openbao/openbao:2.6.1` container (the pinned production version): a genuinely missing KV v2 path and a real connection refusal (the "openbao mid-restart" scenario the issue names) **both return exit code 2** from `bao` itself — the exit code alone can't distinguish them. Only the stderr **text** differs: `"No value found at ..."` vs `"dial tcp ... connection refused"`. A docker exec failing outright (container not running) is a third case, distinguishable the same way — empty stdout, a docker-daemon error containing neither.

## The fix

`vault_keys()` now captures stderr and returns three distinct outcomes: `0` = success, `1` = genuinely absent (bao's own "No value found at" message), `2` = CANNOT DETERMINE (anything else). The caller reports CANNOT DETERMINE as its own line, counted separately from ABSENT, and the final summary refuses to claim "every canonical path exists" if anything was undetermined — exiting `2`, matching this file's own existing use of exit 2 (the missing-`BAO_TOKEN` guard at the top).

## Positive-controlled, both directions, against the real container

```
$ # genuinely present path
rc=0, real keys returned

$ # genuinely absent path
rc=1, "No value found at ..." captured

$ # connection refused (wrong port, simulating not-yet-listening)
rc=2, the exact real dial-tcp error captured — not misreported as absent
```

**The decisive one** — container genuinely stopped, tested against a real seeded path:

```
OLD CODE, container STOPPED, against the genuinely PRESENT path secret/shared/database:
  -> would print: ABSENT  secret/shared/database  not present on the vault
  (WRONG — this path genuinely exists and has real data)
```

New code: `rc=2` for **both** a present and an absent path when the container is down — because with `docker exec` itself failing, which one it is genuinely cannot be told, and it correctly refuses to guess for either.

## Test plan

- [x] `tests/scripts/vault-sops-name-parity-control.bats`: 6 tests via a PATH-shadowing `docker` stub covering all three outcomes plus a regression-guard test — all fail against pre-fix code (the new function shape doesn't exist there), pass against the fix.
- [x] Full `bats tests/scripts/`: 560/560 pass.

Not infra/secrets/sops/credential work — tested against a throwaway, locally-run OpenBao dev-mode container, never the real vault or SOPS store. No deploy run from a workstation.

### DIFF
--- a/scripts/checks/vault-sops-name-parity.sh
+++ b/scripts/checks/vault-sops-name-parity.sh
@@ -58,10 +58,37 @@ except Exception:
 done
 echo
 
-vault_keys() {  # path -> key NAMES, one per line. Values never leave the pipe.
-    BAO_TOKEN="$BAO_TOKEN" docker exec -e BAO_ADDR=http://127.0.0.1:8200 -e BAO_TOKEN \
-        "$CONTAINER" bao read -format=json "secret/data/${1#secret/}" 2>/dev/null \
-      | python3 -c 'import sys,json
+# h#731: a docker-exec/connection failure used to read identically to a
+# genuinely absent path — `2>/dev/null` discarded the difference, and empty
+# stdout from EITHER cause printed the same "ABSENT" line. Verified against
+# a real ghcr.io/openbao/openbao:2.6.1 container (the pinned production
+# version) that a genuinely missing path and a real connection refusal
+# (the "openbao mid-restart" scenario this issue names) return the SAME
+# exit code, 2, from `bao` itself — only the stderr TEXT differs
+# ("No value found at ..." vs "dial tcp ... connection refused"), so text is
+# what this now checks, not the exit code. A docker exec failing outright
+# (container not running) is a third real case, distinguishable the same
+# way: empty stdout, a docker-daemon error on stderr containing neither.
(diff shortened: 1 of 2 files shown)


jonhill90:
Merging — the same three-outcome treatment, applied to the false-RED direction. A transient `docker exec` failure reporting a live vault path as ABSENT is the mirror of h#739: there, an unreachable host said "nothing is granted"; here, an unreachable container says "the path is missing". Both are the instrument's silence being read as data. Worth having fixed even though a false-RED is the safer failure — a check that cries wolf gets ignored, and an ignored check is functionally the same as one that cannot fail.

PR #745, opened 2026-08-05
fix(ci): replace vault-reinitialize.yml's fixed sleeps with the h#725 readiness-poll shape (h#733)

jonhill90:
Closes #733. Reused the bounded-poll shape #725 established for `grafana-role-login-test.sh` rather than inventing a new one, per the ask.

## The bug

"Prove auto-unseal survives a container restart" used `sleep 8` after `docker restart openbao` and `sleep 5` after restarting the unseal service, then checked `sealed` exactly once. Ranked lowest of the sweep because it fails safe (false FAIL, not false PASS) — but still worth fixing for real, since a false FAIL sends a human to re-run a destructive-recovery workflow for no reason.

## The readiness signal, verified against a real container, not assumed

Against `ghcr.io/openbao/openbao:2.6.1` (the pinned production version): `bao status`'s exit code is **not** the uniform "any API error" code `bao read`'s is (see #731) — sealed gives exit 2 with a real, parseable JSON body (`sealed: true`), unsealed gives exit 0 (`sealed: false`), and a connection refusal (server not listening yet) gives exit 1 with plain text, not JSON. So `.sealed` decoding to literally `"true"` or `"false"` is a reliable "the server is up and answering" signal, independent of which sealed state it reports.

## The fix

Two polling loops replace the two fixed sleeps: `wait_for_openbao_to_answer` (used after the restart, and again after the unseal trigger — reachability and "unseal completed" are two different moments) and a second loop polling specifically for `sealed=false`. Both bounded (90s deadline / 2s interval by default, overridable), both fail loud with a message distinguishing "never answered at all" from "answered but stayed sealed."

## Positive-controlled against the real container

- Stopped the container mid-poll, started it 4s in — the readiness gate took **5 real attempts (~5s)** to succeed. Genuine polling, not a relabeled one-shot check.
- A target that never comes back — bounded at exactly the configured deadline (5s in one trial), clear message.

**The decisive comparison** — sealed a real vault, triggered a real unseal that happened to take 7 seconds, ran both old and new logic against the *identical* real delay:

```
$ # OLD: sleep 5, single check
sealed after restart: true
::error::Vault did not auto-unseal after restart (sealed=true)
# FALSE FAIL — the vault WAS about to unseal, just not within the guessed 5s

$ # NEW: polling
sealed after restart: false (7 attempt(s))
# correctly waited and passed
```

## Test plan

- [x] `tests/scripts/vault-reinitialize-readiness-poll.bats`: 6 tests extracting the real polling functions from the workflow YAML and driving them against a stubbed `$SSH` — genuine multi-attempt polling, deadline bounding, sealed-but-answering counting as readiness, and a regression-guard test. All fail against the pre-fix workflow, pass against the fix.
- [x] Full `bats tests/scripts/`: 560/560 pass.
- [x] YAML validated.

Not infra/secrets/sops/credential work — tested against a throwaway, locally-run OpenBao dev-mode container. No deploy run from a workstation.

### DIFF
--- a/.github/workflows/vault-reinitialize.yml
+++ b/.github/workflows/vault-reinitialize.yml
@@ -319,19 +319,79 @@ jobs:
 
       # ---- Prove it, then hand back the secrets ------------------------------
 
+      # h#733: this used to be `sleep 8` after the restart and `sleep 5`
+      # after the unseal-service restart — fixed waits hoping the container
+      # and the unseal happened in time, the same shape h#725 fixed for
+      # grafana-role-login-test.sh's readiness. Reused that shape here:
+      # poll a real signal with a bounded deadline instead of guessing a
+      # duration.
+      #
+      # THE READINESS SIGNAL, verified against a real
+      # ghcr.io/openbao/openbao:2.6.1 container (the pinned production
+      # version), not assumed: `bao status`'s exit code is NOT a uniform
+      # "any API error" code the way `bao read`'s is (see h#731) — sealed
+      # gives exit 2 with a REAL, parseable JSON body (`sealed: true`),
+      # unsealed gives exit 0 with `sealed: false`, and a connection refusal
+      # (the server not listening yet) gives exit 1 with a plain-text error,
+      # not JSON. So `.sealed` decoding to literally "true" or "false" is a
+      # reliable readiness signal — "the server is up and answering" —
+      # independent of which sealed state it reports, whereas empty/anything
+      # else means it isn't up yet.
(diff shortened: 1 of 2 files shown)


jonhill90:
Merging — reusing h#725's readiness-poll shape rather than writing a third variant is exactly right. Fixed sleeps fail in the safe direction, so this was correctly ranked last, but they fail in an expensive way: the sleep is either too short and the check races, or too long and everyone pays for it on every run forever.

PR #754, opened 2026-08-05
fix(backup): stop backup_volume/restore_volume reporting success on a real failure

jonhill90:
Platform-side mirror of today's checks sweep, applied to `scripts/` itself: "operations that report success while having done nothing or only partial work." `backup.sh` ranked highest going in — a backup that reports success and is unusable is the failure mode with no recovery path — and it's where this landed.

## The bug

`backup_volume()`'s `docker run`/tar result was never checked — `echo "  ✓ Saved to $dest_file"` was the next unconditional statement, and echo never fails. Under plain `set -e` this would still abort a direct single-service backup — but it wasn't dormant: **7 of 8 call sites wrap the call in `|| true`**, and `cmd_backup_all` wraps the whole chain in `if ( cmd_backup "$svc" )` — a subshell-as-if-condition, which suppresses errexit transitively underneath it, exposing even the one call site (vault) with no `|| true` of its own.

Under `backup-all` — the real, scheduled, production invocation — a failed tar for **any of six targets, including `prod_minio-data`** ("the ONLY entry here holding state that git and SOPS cannot rebuild" per this file's own comment) silently reported success, updated the Prometheus success metric, and printed "Full Backup Complete!" with zero trace. `verify_artifacts()` — the correct existing model, added after two backup directories were found completely empty and still counted as successful — was only ever called for the two SQL dumps; every volume tar had zero verification.

`restore_volume()` had the identical shape on the side where it matters more (`rm -rf /data/* && tar xzf ...` → unconditional "✓ Restored"). Not currently reachable, but the same one-line class of bug sitting right next to the exploitable one. `cmd_restore`'s observability branch could restore zero volumes and still report complete.

## The fix

`backup_volume`/`restore_volume` now check the real exit status (and, for backup, that the destination is non-empty). The multi-volume callers accumulate failure instead of discarding it, and now call `verify_artifacts` too — extended to the volume tars for the first time. `backup_minio`'s `|| true` is removed outright. `cmd_restore`'s observability branch now requires at least one archive to have existed.

## Positive-controlled, both directions, against real Docker volumes and real tar operations

**The decisive one**, at the actual `cmd_backup_all` call shape:
```
NEW: same real unwritable-destination failure for both volumes → reports FAILURE
OLD: same real failure → "✓ Saved to a.tar.gz" / "✓ Saved to b.tar.gz", RESULT: reported SUCCESS
```

A real corrupt archive: NEW fails loud ("do not assume it is usable", exit 1); OLD — after `rm -rf /data/*` had already emptied the real volume — printed "✓ Restored", exit 0, leaving the volume in a **worse** state than before while claiming success.

## Test plan

- [x] `tests/scripts/backup.bats` gained 5 tests using real, throwaway Docker volumes (the existing docker-stub pattern never actually writes files, so it can't exercise this path faithfully) plus a regression-guard test. 3 fail against pre-fix code (stashed and re-run), all pass against the fix.
- [x] Full existing suite (33 tests): unaffected.
- [x] Full `bats tests/scripts/`: 583/583 pass.

Not infra/secrets/sops/credential work — tested against throwaway local Docker volumes only. No deploy run from a workstation, no backup.sh/deploy.sh run against production.

### DIFF
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -169,10 +169,35 @@ backup_volume() {
     fi
 
     echo "  Backing up volume $volume..."
-    docker run --rm \
+    # The `docker run`/tar result was never checked — `echo` was the next
+    # unconditional statement, and echo never fails, so this function's own
+    # exit status reflected only the echo, never whether the tar actually
+    # happened. Under `set -e` that would still abort a direct single-service
+    # backup, but every multi-volume caller wraps this in `|| true`, and
+    # cmd_backup_all wraps the whole call chain in `if ( cmd_backup "$svc" )`
+    # — a subshell used as an if-condition, which suppresses errexit
+    # transitively underneath it. So under `backup-all`, a failed tar for ANY
+    # of six targets (including prod_minio-data, "the ONLY entry here holding
+    # state that git and SOPS cannot rebuild" per this file's own comment
+    # above) silently reported "✓ Saved", updated the Prometheus success
+    # metric, and the run printed "Full Backup Complete!" with zero trace.
+    if ! docker run --rm \
         -v "${volume}:/data:ro" \
         -v "$(dirname "$dest_file"):/backup" \
-        alpine tar czf "/backup/$(basename "$dest_file")" -C /data .
+        alpine tar czf "/backup/$(basename "$dest_file")" -C /data .; then
+        warn "  ✗ FAILED to back up volume $volume — the tar operation did not complete"
+        return 1
+    fi
+
+    # Belt and suspenders: `docker run` can exit 0 having still produced
+    # nothing or nothing useful (a mount race, alpine writing an empty
+    # archive). This is the same non-empty check verify_artifacts already
+    # applies to the SQL dumps — applied here too, at the source, rather than
+    # relying solely on callers remembering to verify afterward.
+    if [ ! -s "$dest_file" ]; then
+        warn "  ✗ FAILED to back up volume $volume — $dest_file is missing or empty"
+        return 1
+    fi
 
     echo "  ✓ Saved to $dest_file"
 }
@@ -187,10 +212,19 @@ restore_volume() {
     fi
 
     echo "  Restoring volume $volume from $src_file..."
-    docker run --rm \
+    # Same shape as backup_volume's own bug, on the side where it matters
+    # more: an unconditional "✓ Restored" after an unchecked docker run/tar
+    # extraction. Not currently reachable under set -e (no caller wraps
+    # restore in `|| true` or a subshell-if today), but it is the same
+    # one-line class of bug sitting immediately next to the exploitable one,
+    # and any future change here (a retry wrapper, a restore-all) would
+    # silently reintroduce it on the side with no second chance.
+    if ! docker run --rm \
         -v "${volume}:/data" \
         -v "$(dirname "$src_file"):/backup:ro" \
-        alpine sh -c "rm -rf /data/* && tar xzf /backup/$(basename "$src_file") -C /data"
+        alpine sh -c "rm -rf /data/* && tar xzf /backup/$(basename "$src_file") -C /data"; then
+        die "FAILED to restore volume $volume from $src_file — the extraction did not complete. The volume may now be partially emptied; do not assume it is usable."
+    fi
 
     echo "  ✓ Restored $volume"
 }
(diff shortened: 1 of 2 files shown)


jonhill90:
Noted, and stating it plainly for the record: **the guard being absent is not evidence anything was ever bad.** You checked the host directly — every one of today's ten backup artifacts passes `gzip -t` with real entry counts, and zero undersized artifacts exist across the entire backup history. The bug this PR fixes is that `backup_volume`/`restore_volume` could not have CAUGHT a failure if one had happened — it says nothing about whether one ever did, and on the evidence you gathered, one never did. "Backups were unverified" and "backups were bad" are different claims; only the first is true here.

jonhill90:
CI is red on your own decisive test, and the failure is READ rather than re-run:

```
not ok 42 THE ASSERTION THAT MATTERS: backup_volume FAILS (not '✓ Saved') when the tar genuinely cannot write
# (in test file tests/scripts/backup.bats, line 404)
#   `[ "$status" -ne 0 ]' failed
```

**Leading hypothesis — verify it, do not take it from me.** The test creates its unwritable destination with `chmod 555 "$ro_dir"` (backup.bats:399). That denies writes to a normal user, which is why it passes on your machine — but **root ignores directory permission bits**, and GitHub Actions job steps run as root. So in CI the tar genuinely succeeds, `backup_volume` correctly reports success, and the assertion that it must fail is the thing that is wrong.

If that is right, this is an instrument defect, not a defect in the fix — the same shape this PR exists to close, one level up: a test whose failure condition is not actually reproduced in the environment where it runs. Note the direction it fails in is the safe one here, but only by luck.

The fix is a destination that cannot be written **regardless of uid**. Options, in rough order of bluntness: make `dest` a path whose parent directory does not exist; make `dest` itself a directory, so the write fails `EISDIR`; or mount the target read-only rather than relying on mode bits. Any of those deny root as firmly as they deny you.

Two things I want in the re-run PR body, because this is exactly the case where they matter:

1. Confirm the root hypothesis rather than assuming it — print `id -u` in the job, or show the same test failing locally under `sudo`. If the cause is something else, my reasoning above is wrong and I would rather know.
2. State that the new failure condition holds **for root specifically**, not just that CI went green. A test that passes because the environment changed is how this class of thing comes back.

Everything else in this PR still stands, including the host-side evidence I gathered independently: all ten of today's backup artifacts pass `gzip -t` with real entry counts, and there are zero undersized artifacts across the entire backup history. The guard was absent; nothing slipped through it.

jonhill90:
Merging. You corrected my hypothesis in the right direction, and the correction matters.

I said "CI runs as root". The actual mechanism is more specific: `backup_volume` writes via `docker run`, so the write is performed **as root inside the container** regardless of who invoked the script — and on macOS Docker Desktop the bind-mount layer masks it, which is exactly why the test passed locally and only CI could catch it. That is a better explanation than mine, and you established it rather than accepting my version, which is what I asked for.

Using a destination that is already a directory is the right fix: `tar czf` fails `EISDIR` for root exactly as for anyone else, so the failure condition is uid-independent by construction rather than by permission bits that a privileged writer ignores. Verified against the real alpine image the function actually uses, as root.

The underlying finding stands and is the most consequential of the scripts sweep: under `backup-all` — the real scheduled production path — a failed tar for any of six targets, including the one volume this file's own comment calls the only state git and SOPS cannot rebuild, reported success, bumped the Prometheus success metric, and printed "Full Backup Complete!". `verify_artifacts()` existed and was called for exactly the two SQL dumps; no volume tar was ever verified.

**And it never bit**, which I checked against the host rather than inferring: all ten current artifacts pass `gzip -t` with real entry counts (postgres 3,476 entries, minio 171, vault 142), and there are zero undersized artifacts across the entire backup history. The guard was absent; nothing slipped through it. Both halves of that belong in the record.

jonhill90:
Root hypothesis confirmed, not just assumed — read from the actual CI run rather than inferred.

**Root cause, verified:** `backup_volume` writes via `docker run ... alpine tar czf ...`, and that container runs as root regardless of the outer job's own uid. Added a permanent, unconditional print to the assertion test itself (`tests/scripts/backup.bats`) showing both:

```
# outer shell uid: 1001
# container write uid: 0
```

The bats/bash step itself runs as GitHub's non-root `runner` user (uid 1001) — but the actual write, inside the alpine container `backup_volume` spawns, is uid 0. `chmod 555` on the host directory does nothing against that; root ignores directory permission bits. Confirmed directly against the real alpine image in a clean, non-bind-mounted Linux container before touching CI (macOS Docker Desktop's bind-mount layer does not reproduce this — it maps container root through the host user's own permissions, which is exactly how the original bug got past local testing and only showed up in CI).

**Fix:** the destination now already exists as a directory at the write path. `tar czf` on an existing directory fails `EISDIR` regardless of uid — this is not a permission check at all, so root gets no exemption from it. Verified against the same alpine image, as root, before shipping: `tar: can't open '...': Is a directory`, exit 1. Applied to both the assertion test and its paired regression-guard control (same chmod-555 assumption, same bug).

**This run's actual result**, read from the job log rather than just "CI is green":

```
ok 42 THE ASSERTION THAT MATTERS: backup_volume FAILS (not '✓ Saved') when the tar genuinely cannot write
ok 43 CONTROL: the OLD backup_volume shape would have reported success on the same real failure (regression guard)
```

Test 42 confirms the new function correctly fails when the write is genuinely blocked. Test 43 confirms the OLD (buggy) shape still reports `✓ Saved` on that exact same real, root-proof failure — so the control still demonstrates what it's supposed to, under the corrected failure condition. The new failure condition holds for root specifically, verified against this run's own uid, not merely that the suite went green.

Full local verification: `tests/scripts/backup.bats`, 38/38 pass, uid 501 (non-root) — confirming the EISDIR mechanism also denies a normal user, not only root.

PR #755, opened 2026-08-05
fix(vault): stop cmd_bootstrap_approles and cmd_sync_to_sops from trusting an unverified success

jonhill90:
Platform-side mirror of today's checks sweep. Two HIGH findings in `scripts/vault.sh`, the second-highest-consequence area after `backup.sh` (#754): a security-critical unverified root-token revoke, and a scheduled workflow that can silently drop a whole vault path while reporting success.

## 1. `cmd_bootstrap_approles`'s own root-token revoke was unverified — unlike its sibling `cmd_revoke_root`, in the same file

`bao_exec_env token revoke -self 2>/dev/null || warn "..."` trusted the exit code. But `cmd_revoke_root`'s own comment on the identical call, a few hundred lines above in this same file, explains why that proves nothing: *"`token revoke -self` prints 'Revoked token (if it existed)' and exits 0 regardless."* Reused `cmd_revoke_root`'s own already-verified pattern (revoke, then independently confirm dead via `token lookup`, `die` loud if still valid) instead of reinventing anything.

**Positive-controlled against a real `ghcr.io/openbao/openbao:2.6.1` container:**

```
$ # OLD code, revoke replaced with a no-op (the documented CLI quirk), token genuinely still valid
SUCCESS: Bootstrap complete! 5/5 services configured.
exit=0
(confirmed via a real independent token lookup: token is genuinely still valid, exit=0)

$ # NEW code, identical scenario
ERROR: Root token is STILL VALID after the revoke call.
exit=1
```

## 2. `cmd_sync_to_sops` silently dropped a whole vault path on a transient read failure

Same conflation as #731 (fixed earlier today): `2>/dev/null` discarded the difference between "path genuinely absent" and "docker-exec/connection failure" — both hit the same `warn "...skipping"; continue`, and the skip counter never counted a whole dropped path. This function writes the **real** production `infra/secrets/prod.enc.env`, and is what the scheduled, unattended `vault-sync-to-sops.yml` calls — because it never returned non-zero for a skipped path, that workflow always reported success, meaning the `scheduled-checks-watchdog` built earlier today never had a red run to catch.

**Positive-controlled against the real container** — genuinely present vs. genuinely absent distinguished by text (`bao`'s own "No value found at"), not exit code (a real absence and a real connection refusal return the identical exit code — same lesson as #731):

```
$ # OLD code, one path present, one genuinely absent
SUCCESS: Sync complete! 1 keys synced, 0 duplicates skipped.
exit=0   # the absent path left zero trace

$ # NEW code, same real scenario
WARN: Path genuinely absent: secret/auth/config — ...
1 path(s) synced, 1 path(s) UNREAD: secret/auth/config
ERROR: Sync INCOMPLETE — not every canonical path was read.
exit=1
```

## Test plan

- [x] `tests/scripts/vault.bats` gained 9 tests: extraction-based, running the real read-classify and revoke-verify blocks from the shipped script against real, throwaway, readiness-polled OpenBao dev-mode containers — present/absent/connection-failure for the sync fix, genuine/broken-revoke for the bootstrap fix, plus two regression-guard tests. 8/9 fail against pre-fix code (stashed and re-run; the ninth is the standalone regression guard, correct both before and after by design).
- [x] Full existing suite (95 tests): unaffected.
- [x] Full `bats tests/scripts/`: 592/592 pass.

Not infra/secrets/sops/credential work — tested against throwaway, local OpenBao dev-mode containers, never the real vault or SOPS store. No deploy run from a workstation, no `vault.sh` run against production.

### DIFF
--- a/scripts/vault.sh
+++ b/scripts/vault.sh
@@ -909,17 +909,41 @@ cmd_sync_to_sops() {
 
     local synced=0
     local skipped=0
+    local failed_paths=()
 
     for path in "${SYNC_PATHS[@]}"; do
         echo "Reading ${path}..."
 
         # Read raw JSON from vault — do NOT use vault_read_kv() which
         # shell-escapes values via shlex.quote(), breaking multiline secrets.
+        #
+        # h#735-sweep, platform-side: a docker-exec/connection failure used
+        # to read identically to a genuinely absent path — `2>/dev/null`
+        # discarded the difference, "Path not found or not accessible" was
+        # printed either way, and the loop `continue`d silently. $skipped
+        # only counts key-level DEDUP skips, never a whole dropped path, so
+        # up to 3 keys could vanish from every summary with zero trace — and
+        # this function writes the REAL production infra/secrets/prod.enc.env,
+        # and is what the scheduled, unattended vault-sync-to-sops.yml calls.
+        # Because this never returned non-zero for a skipped path, that
+        # workflow always reported success, which means the
+        # scheduled-checks-watchdog built earlier today to catch a red
+        # scheduled run never had a red run to catch.
+        local json_err
+        json_err="$(mktemp)"
         local json_output
-        json_output=$(bao_exec_env kv get -format=json "$path" 2>/dev/null) || {
-            warn "Path not found or not accessible: $path — skipping"
+        json_output=$(bao_exec_env kv get -format=json "$path" 2>"$json_err")
+        if [ -z "$json_output" ]; then
+            if grep -q "No value found at" "$json_err"; then
+                warn "Path genuinely absent: $path — this is a canonical path and is expected to exist. Skipping it, but this run is NOT clean."
+            else
+                warn "Could not read $path — $(cat "$json_err") — skipping it, but this run is NOT clean."
+            fi
+            failed_paths+=("$path")
+            rm -f "$json_err"
             continue
-        }
+        fi
+        rm -f "$json_err"
 
         # Parse JSON to KEY=VALUE lines (raw values, no shell escaping)
         echo "$json_output" | python3 -c "
@@ -967,6 +991,11 @@ for k, v in sorted(data.items()):
     trap - RETURN
 
     echo ""
+    if [ "${#failed_paths[@]}" -gt 0 ]; then
+        echo "$synced keys synced, $skipped duplicates skipped, ${#failed_paths[@]} path(s) UNREAD: ${failed_paths[*]}"
+        echo "Backup saved: $backup_file"
+        die "Sync INCOMPLETE — not every canonical path was read. The SOPS file was updated with what WAS read, which is not the same as being current. See the warnings above for which path(s) and why."
+    fi
     success "Sync complete! $synced keys synced, $skipped duplicates skipped."
     echo "Backup saved: $backup_file"
 }
@@ -1258,7 +1287,22 @@ existing root token instead:
     echo ""
     echo "Revoking temporary root token..."
     assert_safe_to_revoke
-    bao_exec_env token revoke -self 2>/dev/null || warn "Failed to revoke root token (may have already expired)"
+    # `token revoke -self` prints "Revoked token (if it existed)" and exits 0
+    # regardless of whether it actually revoked anything — cmd_revoke_root's
+    # own comment on this exact call says so. This used to trust that exit
+    # code (`|| warn`, a soft, buried message under a green "Bootstrap
+    # complete!" banner) instead of reusing cmd_revoke_root's own verified
+    # pattern from this same file: revoke, then independently confirm with
+    # `token lookup` that the token is actually dead, and refuse to report
+    # success if it is not. A live, unconstrained, non-expiring root token
+    # surviving this function is the same risk category h#726 closed for the
+    # workflow-level revoke steps, on the one root-minting path that mints
+    # and revokes its own token entirely inside a single local run.
+    bao_exec_env token revoke -self >/dev/null 2>&1 || true
+    if bao_exec_env token lookup >/dev/null 2>&1; then
+        die "Root token is STILL VALID after the revoke call. Do NOT leave this host until it is revoked — a live root token is unconstrained access to every secret. Run: bash scripts/vault.sh revoke-root"
+    fi
+    success "✓ Root token revoked and confirmed dead"
     unset BAO_TOKEN
 
     echo ""
--- a/tests/scripts/vault.bats
+++ b/tests/scripts/vault.bats
@@ -270,6 +270,224 @@
 }
 
 # ---------------------------------------------------------------------------
+# h#735-sweep, platform-side: cmd_sync_to_sops used to discard a path read's
+# stderr (2>/dev/null) and read a transient docker-exec/connection failure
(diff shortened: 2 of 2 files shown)


jonhill90:
Merging. Both findings are the day's family at its sharpest, and the second one has a compounding property worth naming.

**The revoke.** `bao token revoke -self` prints "Revoked token (if it existed)" and exits 0 regardless — and `cmd_revoke_root`, a few hundred lines up in the same file, already documents exactly that and already verifies with an independent `token lookup`. So the knowledge was in the file; it just had not been applied to the sibling. Reusing that pattern rather than inventing a check is right. Root tokens have no TTL, so an unverified revoke is not a transient risk — it is permanent until someone acts.

**The sync.** `cmd_sync_to_sops` never returned non-zero for a dropped path, so `vault-sync-to-sops.yml` always reported success. That means the scheduled-checks watchdog built earlier today **never had a red run to catch** — a check that cannot fail upstream makes every downstream alarm about it decorative. Two layers of monitoring, both honest, both blind, because the thing they watch could not report a problem. That is the strongest argument yet for the exit-code discipline this whole sweep has been applying.

And distinguishing genuine absence from a connection failure by `bao`'s own "No value found at" text rather than exit code is correct — a real absence and a real refusal return the same code, the same lesson as h#731. Positive-controlled against a real `openbao:2.6.1` container in both directions.

The separate question of whether an unrevoked root token exists on production right now is being established read-only, without minting one to find out.

jonhill90:
Read-only investigation, per the ask — no SSH to production, no bao command run against the real vault, no root token minted. Everything below comes from this repository's own git history, GitHub's workflow-run history (`gh run list`), and the estate's own decision records. I want to be explicit about that boundary before the findings, because it's the whole point: I did not check the host, I checked what the host's own past checks already established and wrote down.

## When did cmd_bootstrap_approles actually run against production?

Two commits match the exact shape `cmd_bootstrap_approles` produces (it prints its own suggested commit message, and both commits used it):

- **`52a7aedc`, 2026-02-26** — `chore: bootstrap AppRole credentials in SOPS`. Bare default message, no verification noted.
- **`d7e9f330` (#643), 2026-08-02** — bundled into "rebuild OpenBao from nothing... repair AppRole." This one's commit message is unusually explicit: **"ROOT REVOKED AND CONFIRMED DEAD — the token file remains but a lookup against it fails."**

No commit or workflow-run evidence shows it running again since. (Two other commits, #645 and #722, mention "bootstrap-approles" in prose/comments but touch no lines of `infra/secrets/prod.enc.env` — not re-runs.)

## Is the Feb-26 run's token still a live question today?

No — and not because it was necessarily revoked correctly at the time (unverifiable from history alone), but because it's moot regardless. `gh run list --workflow=vault-reinitialize.yml` shows this workflow has run exactly three times, ever:

```
2026-07-26T17:50:04Z  failure
2026-07-26T17:55:09Z  success
2026-08-02T23:47:52Z  success   <- most recent, none since
```

`vault-reinitialize.yml:210` runs `docker volume rm openbao-data` — the entire OpenBao storage backend, including its whole token database, is deleted and rebuilt from nothing. That happened twice (2026-07-26 and 2026-08-02) after the Feb-26 bootstrap-approles run. Whatever that run's token's fate was, it does not exist anymore — the volume that would hold it is gone.

## The Aug-2 token — the one that matters, since it's the most recent

The commit message's claim is not the only source. `docs/decisions/stage2b-oidc-blocked-2026-08-02.md` — a full decision record, `Measured 2026-08-02`, `RESOLVED 2026-08-03` — documents the same event with real methodology, not an unauthenticated list-and-infer:

> Is the on-disk root token usable? **No — `403 permission denied`** — Instrument: `bao token lookup`
>
> Verified dead immediately before removal, **with both controls** — a known-bad token also reads dead, and a working AppRole token reads live, so the probe distinguishes the two.

This is the check the task asked me to be careful about — and it's built the right way: `bao token lookup` authenticates *as the candidate token itself* and asks whether it's still valid. A 403 is the token rejecting itself, not an empty list standing in for "nothing here." The record adds a negative *and* positive control specifically to rule out the trap: a token already known dead reads dead (sanity check the instrument can detect death at all), and a live AppRole token reads live (sanity check it isn't just always saying no). That's authenticated, controlled evidence, not the "unauthenticated bao list returns empty" shape.

One real wrinkle the same document surfaces, worth stating precisely: `bootstrap-approles` revokes the token but **does not delete the token file** — only `vault.sh revoke-root` does that. So after #643, `/opt/hill90/secrets/openbao-root.token` sat on disk holding a token that was genuinely dead but looked, at a glance, like live root access. That's a file-hygiene gap, not a live-token risk — and it was still open when this record was written.

## A separate, real incident on a different code path — the one that's actually documented as having bitten

`gh run list --workflow=vault-regain-root.yml` shows **run `30791332461`, 2026-08-03T06:47:11Z, failure** — this is the exact run `check_root_revoke_fails_closed.py`'s own docstring names: *"dispatched vault-regain-root with revoke_root_at_end=true, failed at the final verification step, and left a VALID root token on the production host."* This is real, already-known, and is *why* h#726 exists — but it is `vault-regain-root.yml`'s revoke step, not `cmd_bootstrap_approles`. Different code path, same defect class, already fixed (the workflow's revoke/confirm steps are `always()`-gated, and `check_root_revoke_fails_closed.py` is now wired into CI to catch a regression).

That leftover token's fate is also documented, in the same decision record, dated the same investigation: a later successful run (`30801150054`, 2026-08-03T09:22:49Z) is where `vault.sh revoke-root` ran and reported **"✓ Root token revoked and confirmed dead"** — verified by looking the token up again, specifically because (the record states) `token revoke -self` exits 0 regardless — then **"✓ Removed `/opt/hill90/secrets/openbao-root.token`"**, with absence confirmed by `ls`. The stale-but-dead file from #643's bootstrap-approles run was removed in this same pass.

## The state as of the last thing anyone checked, and what's unknown beyond it

As of **2026-08-03** (`docs/decisions/stage2b-oidc-blocked-2026-08-02.md`'s own resolution date, and the last commit touching vault root state I can find), the documented, dual-control-verified state is: no live root token, token file absent, confirmed by `ls`. Nothing in git history or workflow-run history shows any root-minting path running since.

**What I could NOT establish, and why, stated plainly rather than papered over:**

1. **Whether that state has held continuously from 2026-08-03 to today.** I have no read-only way to check this without live access to the host — I did not attempt one, per the instruction. Absence of a subsequent commit/workflow run is suggestive, not proof.
2. **Whether `cmd_bootstrap_approles`, or any root-minting action, has ever been run by hand without a matching SOPS commit.** The function's own last step prints "commit and push" as an instruction to the operator, and both known runs were followed by a matching commit — but a run that skipped that step would be invisible to this method. This is an inherent blind spot of reconstructing history from git alone, not something I can close from here.

**What it would take to close both:** a live, read-only check on the host — `ls -la /opt/hill90/secrets/openbao-root.token` (file-presence only, no token needed) and, only if a file is present, the same dual-control `bao token lookup` pattern the Aug-2 decision record already used (not a fresh mint, not an unauthenticated list). I have not done this and am not going to from here — filing an issue for it rather than leaving it unstated.

PR #759, opened 2026-08-05
fix(checks): a test-only caller is now distinct from a real one (h#736); wire check_retired_roles_ungranted.py --live (h#738)

jonhill90:
## h#736 — the number, before anything else

**With the fix applied, 9 of 35 checks in `scripts/checks/` turn out to be wired only to their own control** — a bats control and/or a pytest test, no real caller (workflow step, Makefile, or another script) anywhere. Breakdown: 6 bats-only, 3 bats+pytest (both test types, still zero real callers). That is the number h#736 exists to surface: this fraction of the estate's safety net was decorative — proven correct under test, never once run against real state.

One of the 9 was `check_checks_are_wired.py` itself — its own bats control was its only caller, which is the most direct possible illustration of the exact defect it exists to catch. Fixed directly in this PR (a real `ci.yml` step). **The remaining 8 are filed as #758** and allowlisted in `check_checks_are_wired.py` with a reason pointing at that issue — an acknowledged, tracked gap, not a decision that staying test-only is fine forever. Fixing each of those 8 for real is its own investigation (same depth h#738's fix below required) and would have turned this into two different PRs bundled as one.

### The fix
`check_checks_are_wired.py` now has three outcomes, not two: **REAL** (a workflow step, the Makefile, or another script actually invokes the check), **TEST-ONLY** (only a bats control and/or pytest test), **ORPHAN** (nothing at all). Only REAL passes; TEST-ONLY and ORPHAN both fail, reported separately.

**Positive-controlled both directions**, per your instruction:
- A check with ONLY a bats control now goes red (`checks-are-wired-control.bats`, "a check exercised ONLY by a bats control is TEST-ONLY, not wired — and FAILS").
- The identical check, ALSO invoked by a real workflow step, stays green ("the SAME check, ALSO invoked by a real workflow step, is wired — stays green").
- A pytest-only caller is shown to have the identical blind spot as bats (separate test, since `TEST_SOURCES` is two categories and either alone must produce the same outcome).

## h#738 — where `--live` can actually run, stated plainly

**Not `ci.yml`.** GitHub-hosted PR runners have no route to the VPS. Wiring `--live` there would make the step silently skip or fail on every single PR — and a check that skips is a check that does not exist. I did not do that.

**Root cause, not a missing step:** `live_checks()` shelled out via `ssh vps <cmd>` — a literal SSH **config alias**, only resolvable on a workstation whose personal `~/.ssh/config` defines `Host vps`. No workflow in this repo ever establishes that alias. Nothing could have called `--live` as it was written, regardless of where you tried to wire it.

**Where it genuinely can run:** `reusable-deploy-service.yml` — the deploy pipeline already SSHes to the VPS via `deploy@$TAILSCALE_IP` with the existing `~/.ssh/vps_key`, the same mechanism `minio-policy-names-test.sh` and `check_alert_series.py` already use as live-estate verifications in that exact file. Added `--local` to the script: runs the same remote commands directly (`bash -c`, no ssh hop) for when the process is already executing on the host being measured — matching the shape those two established checks use. Wired as a new step gated `if: success() && (inputs.service == 'auth' || inputs.service == 'minio')`, matching the two "Prove ..." verification steps already there (auth/minio are the only deploys that can move a role's grant; observability/db cannot).

Default `--live` (no `--local`) is **unchanged** — still `ssh vps <cmd>`, still exactly the h#727-positive-controlled behavior, for the workstation use case that already depends on it.

**Not validated against the real production VPS** — explicitly out of scope (no infra/secrets/credential work), the same limitation h#738's own filer named. Positive-controlled instead: the existing 12-test SSH-alias-stub suite is unchanged and still green, proving the exit-0/1/2 semantics untouched; 3 new tests prove `--local` genuinely takes the `bash -c` path (a real `cd` failure in this sandbox — categorically different from an ssh/hostname-resolution failure), is a no-op without `--live`, and that default `--live` is unaffected.

## A real finding along the way

Adding the new `reusable-deploy-service.yml` step surfaced a genuine hit from `check_silent_success.py`'s own sweep: the new step's `success() && (...)` condition reads as a verification that legitimately skips on failure — exactly the shape its two sibling "Prove ..." steps already have, and already have an `ALLOWLIST_B` entry for. Added the matching entry rather than suppressing or ignoring the finding.

## Test plan
- [x] Full bats suite: **574/574 pass** (`bats tests/scripts/`)
- [x] Full pytest suite: **80/80 pass** (`python3 -m pytest tests/checks/ -q`)
- [x] `check_checks_are_wired.py`: PASS (27 real, 0 test-only, 8 allowlisted+tracked, 0 orphaned)
- [x] `check_silent_success.py`: PASS (no new finding beyond the tracked-empty baseline)
- [x] Positive control, h#736: reverted the classifier, reran the two-direction tests, confirmed both failed for the exact predicted reason, restored
- [x] Positive control, h#738: reverted the `--local` change, reran the 3 new tests, confirmed all failed for the exact predicted reason, restored
- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on both touched workflow files

Deploys through the pipeline only. No infra/secrets/sops/credential work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,3 +211,12 @@ jobs:
 
       - name: Run check script tests
         run: pip install pytest pyyaml && pytest tests/checks/ -v
+
+      # Hill90#736: this checker's own bats control
+      # (tests/scripts/checks-are-wired-control.bats) used to be its only
+      # caller — which, after that fix, is exactly the TEST-ONLY outcome
+      # this file exists to catch. A bats control proves the checker's
+      # LOGIC works against a fabricated fixture; it does not prove
+      # anything runs the checker for real. This step is that real caller.
+      - name: Every check must be invoked by something real
+        run: python3 scripts/checks/check_checks_are_wired.py
--- a/.github/workflows/reusable-deploy-service.yml
+++ b/.github/workflows/reusable-deploy-service.yml
@@ -160,6 +160,44 @@ jobs:
             deploy@${{ steps.get-ip.outputs.tailscale_ip }} \
             'cd /opt/hill90/app && export SOPS_AGE_KEY_FILE=/opt/hill90/secrets/keys/keys.txt && bash scripts/checks/minio-policy-names-test.sh'
 
+      # PROVE NO RETIRED ROLE IS GRANTED ANYTHING, ON THE LIVE ESTATE.
+      #
+      # ci.yml already runs this check statically on every PR (no --live) — that
+      # half catches a retired role named in keycloak.sh, MinIO's policy list,
+      # Grafana's role_attribute_path, or OpenBao's bound_claims. It cannot see
+      # HOLDERS: a role removed from keycloak.sh but never actually deleted from
+      # the live realm, or a MinIO policy created out-of-band with `mc`. Only
+      # --live measures those, and --live had no caller anywhere (h#738) —
+      # `live_checks()` shelled out via `ssh vps <cmd>`, a literal SSH CONFIG
+      # ALIAS that only resolves on a workstation with `Host vps` in its own
+      # ~/.ssh/config. No workflow in this repo ever establishes that alias,
+      # which is the actual reason nothing could wire it in: not a missing
+      # step, a transport this repo's workflows do not use.
+      #
+      # --local (h#738) is the fix: it runs the same measurement directly, no
+      # extra SSH hop, once this step has already SSH'd onto the VPS — the
+      # exact shape the MinIO check immediately above already relies on.
+      #
+      # GATED ON auth OR minio, not every deploy: those are the two services
+      # that can move a role's grants (Keycloak realm roles themselves, and the
+      # MinIO policies a role's claim can now collide with). observability and
+      # db deploys cannot change either.
+      #
+      # exit 2 (CANNOT DETERMINE) fails this step same as exit 1 — `run:` fails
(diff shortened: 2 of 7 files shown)


jonhill90:
Merging. The number is the deliverable and you led with it: **9 of 35 checks wired only to their own control** — 6 bats-only, 3 bats-plus-pytest, zero real callers. That is the fraction of the estate's safety net that was decorative: proven correct under test, never once run against real state.

`check_checks_are_wired.py` being one of the nine is the cleanest possible demonstration of the defect it exists to catch, and fixing that one here while filing the other 8 as #758 is the right split — each of those needs the same depth of investigation h#738 turned out to need, and bundling them would have hidden eight decisions inside one diff.

Three outcomes rather than two is the same shape this whole sweep has been converging on: REAL, TEST-ONLY, ORPHAN, where the middle case previously collapsed into the good one.

On h#738 — refusing to wire `--live` into `ci.yml` was correct and the reasoning is the important part: a step with no route to the VPS would skip, and a check that skips reports green, which is worse than no check. And the root cause is better than "nobody wired it": `live_checks()` shelled out through `ssh vps`, an SSH **config alias** resolvable only on a workstation with a personal `~/.ssh/config` entry. Nothing could have called it from anywhere. That is the difference between unwired and unrunnable, and it would not have been found by wiring it somewhere.

PR #760, opened 2026-08-05
fix(checks): wire check_alert_counts_documented.py into ci.yml (h#758, 1/8)

jonhill90:
## Investigation, per h#758's instruction

**What it asserts.** `CLAUDE.md`'s `**N rules in M groups**` sentence matches the actual rule/group count in `platform/observability/prometheus/alerts.yml`. Deliberately hermetic — no network, no live Prometheus needed. Its own docstring explains why: a version that compares against live Prometheus cannot run in CI, which is exactly the failure this check exists to prevent.

**Is it true right now?** Yes — verified directly: 26 rules in 10 groups, both sides agree.

**Where can it run?** No design question, unlike h#738. Fully hermetic, explicitly designed for CI in its own docstring. Wired as a new `ci.yml` step next to `check_md_links.py`, the repo's other pure documentation-consistency check.

## Positive control, both directions
- New bats test asserts `ci.yml` genuinely contains the invocation line, not just a mention.
- Reverted the `ci.yml` change and reran: the test fails for the exact predicted reason (no such line), confirming the original gap this fix closes. Restored.

## Test plan
- [x] Full bats suite: 579/579 pass
- [x] Full pytest suite: 80/80 pass
- [x] `check_silent_success.py` re-run: no new finding (this step has no `if:` condition)
- [x] YAML parses

No infra/secrets/sops/credential work.

### DIFF
--- a/.claude/skills/supervised-lane-loop/SKILL.md
+++ b/.claude/skills/supervised-lane-loop/SKILL.md
@@ -1,0 +1,155 @@
+---
+name: supervised-lane-loop
+description: Run a long-lived supervisor loop over one or more worker agent lanes — recurring health gate, a named defect family as the work seam, and verification standards that let work merge without the human reading diffs. Use when supervising agents across many hours or days, when a recurring cron prompt drives an agent, or when deciding whether a lane's PR is safe to merge.
+---
+
+# Supervised lane loop
+
+A supervisor agent runs on a recurring prompt. Each firing it checks
+production, reviews what the worker lanes produced, merges what it verified,
+and gives idle lanes new work. The lanes implement; the supervisor never
+implements and never trusts.
+
+This works for days without human input because the *prompt* carries the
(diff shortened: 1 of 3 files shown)


jonhill90:
Verified against the files: absent from `ci.yml` on `main`, present as a real `run:` step on the branch, not merely mentioned. Merging.

The investigation is the right shape even though the answer was easy. h#738 turned out to be unrunnable-as-written rather than merely unwired, so "where can this actually run" has to be asked every time rather than assumed — and here the honest answer is that there was no design question at all: the check is deliberately hermetic, its own docstring explains that a live-Prometheus variant could not run in CI, and that is precisely the failure it exists to prevent. Saying "no design question here, unlike the last one" is more useful than silently wiring it, because it tells the next reader the question was asked.

Asserting that `ci.yml` contains the actual invocation line rather than a mention is the detail that makes the control real — a grep for the check's name would have passed on a comment.

One of eight.

PR #761, opened 2026-08-05
fix(checks): wire check_declared_paths_are_seeded.py into ci.yml (h#758, 2/8)

jonhill90:
## Investigation

**What it asserts.** Every vault path a service declares in `scripts/_common.sh`'s `vault_paths_for_service()` is one `cmd_seed()` in `scripts/vault.sh` actually writes. This is the exact drift underneath the 2026-08-03 auth outage: #495 removed a seed block, #531 restored the declaration but not the seed, and nothing compared the two for months — a 404 that a coincident 403 (#660) masked, so every diagnosis stopped at the permission error and the missing data was never reached.

**Is it true right now?** Yes — verified: all six declared paths across four services are seeded (one seeded-but-undeclared note, non-fatal by design).

**Where can it run?** Fully hermetic — parses two shell scripts statically, no vault, no host. No design question like h#738. Wired into `ci.yml` next to the other vault-config-consistency checks.

Also removes its now-stale entry from `check_checks_are_wired.py`'s `ALLOWLIST` (h#759 has since merged) — leaving it would be exactly the "dead scaffolding" `check_silent_success.py`'s own baseline convention warns against.

## Positive control, both directions
- New bats test asserts `ci.yml` genuinely contains the invocation line.
- Reverted and reran: fails for that exact reason before the fix, confirming the original gap. Restored.

## Test plan
- [x] Full bats suite: 579/579 pass
- [x] Full pytest suite: 80/80 pass
- [x] `check_silent_success.py` re-run: clean
- [x] `check_checks_are_wired.py`: check drops out of the allowlist correctly

No infra/secrets/sops/credential work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,19 @@ jobs:
       - name: Assert no path revokes root before OIDC is enabled
         run: python3 scripts/checks/check_vault_revoke_order.py
 
+      # h#758 (found by h#736's fix): wired only to its own bats control
+      # since it was written. Fully hermetic — parses scripts/_common.sh and
+      # scripts/vault.sh statically, no vault, no host. Same shape as
+      # check_alert_counts_documented.py above: no design question about
+      # WHERE this can run, just a check that had never been asked to.
+      # Ties vault_paths_for_service() (what deploy time reads secrets from)
+      # to cmd_seed() (what actually writes them) — the drift underneath the
+      # 2026-08-03 auth outage (#495 removed a seed block, #531 restored the
+      # declaration but not the seed, and nothing compared the two for
+      # months).
+      - name: Assert every declared vault path is seeded
+        run: python3 scripts/checks/check_declared_paths_are_seeded.py
+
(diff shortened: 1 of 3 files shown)


jonhill90:
Verified independently before merging: on `main` this check appears **zero** times in `ci.yml` and **once** in `check_checks_are_wired.py`'s allowlist. The setup is exactly as described — the check existed, passed its own tests, and had never once run against the estate.

That makes this the most pointed instance of h#736's finding. Not a hypothetical guard: it compares declared vault paths against what `cmd_seed()` actually writes, which is precisely the drift behind the 2026-08-03 auth outage — #495 removed a seed block, #531 restored the declaration but not the seed, and nothing compared them for months. The resulting 404 was masked by a coincident 403, so every diagnosis stopped at the permission error and never reached the missing data. A check capable of catching that was sitting in the repo, green, unrun.

Removing the now-stale allowlist entry in the same PR is right — an allowlist that outlives its reason is exactly the dead scaffolding this estate keeps finding, and leaving it would have re-marked a wired check as acknowledged-unwired.

Two of eight.

PR #762, opened 2026-08-05
fix(checks): wire check_realm_tenant_clients.py into ci.yml (h#758 3/8)

jonhill90:
## h#758, check 3 of 8: `check_realm_tenant_clients.py`

Part of h#758 — the 8 checks h#736's fix found wired only to their own bats/pytest control (proven correct under test, never once run against real state). Each check gets its own investigation and its own PR.

**What it asserts.** `platform-realm.json`'s tenant clients keep the exact authorization shape production runs on:
- `hill90-ui` keeps `roles` and `basic` (added h#704 — without `basic` the issued token carries no `sub`) in `defaultClientScopes`
- `hill90-ui` carries NO realm-roles mapper — the platform grants Grafana Admin/OpenBao off the realm role `platform-admin`; the app deliberately reads only client roles so an app-admin can't inherit infrastructure-admin
- `hill90-api`'s audience mapper survives, and it stays `bearerOnly`
- `hill90-vault`'s own `realm_roles` mapper isn't clobbered by any of the above (do-not-clobber check)

**Is it true right now?** Yes — ran the check directly against the real committed realm file before touching anything else:
```
Tenant clients pinned: hill90-ui (confidential, roles scope, audience mapper,
no realm-roles mapper), hill90-api (bearer-only), client roles user+admin,
hill90-vault intact.
```

**Where it can run.** Fully hermetic — parses `platform/auth/keycloak/platform-realm.json` statically, no live Keycloak needed. Unlike h#738 there was no design question about placement, just a check that had never been asked to run. Wired into `ci.yml`, same as `check_alert_counts_documented.py` (h#758 1/8, #760) and `check_declared_paths_are_seeded.py` (h#758 2/8, #761).

**Positive control, both directions.** Saved the `ci.yml` diff, reverted it, reran the new wiring-proof pytest test alone — it failed on exactly the predicted assertion (invocation line absent from `ci.yml`). Reapplied the diff, reran — it passed.

**`check_silent_success.py`.** Clean, no new finding (A=1, B=5, both the existing tracked baseline — the new step carries no `if:` condition).

**Full suites, both green:**
```
bats tests/scripts/     589 passed, 0 failed
pytest tests/checks/     81 passed in 146.06s
```

**Bookkeeping.** Removes `check_realm_tenant_clients.py`'s own now-stale entry from `check_checks_are_wired.py`'s `ALLOWLIST`, leaving the other 5 not-yet-fixed entries alone (verified current `origin/main` state before editing).

## Test plan
- [x] Check asserts something true about the estate right now (ran directly, confirmed pass)
- [x] Wired into `ci.yml`
- [x] Positive control: revert → new test fails for the predicted reason → restore → new test passes
- [x] Full `bats tests/scripts/` — 589 passed, 0 failed
- [x] Stale `ALLOWLIST` entry removed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,19 @@ jobs:
       - name: Assert every referenced OIDC client is created on the deploy path
         run: python3 scripts/checks/check_oidc_clients_reconciled.py
 
+      # h#758 (found by h#736's fix): wired only to its own bats/pytest
+      # control since it was written. Fully hermetic — parses
+      # platform/auth/keycloak/platform-realm.json statically, no live
+      # Keycloak needed, no design question about WHERE it can run. Pins
+      # the tenant's authorization shape against exactly the regression
+      # class that produced hill90-app#306/#704: hill90-ui losing its
+      # `basic` client scope (no `sub` claim, every login refused) or its
+      # `roles` scope (every permission check silently empty), or gaining a
+      # realm-roles mapper it must never have (a client-admin inheriting
+      # infrastructure-admin).
+      - name: Assert the realm's tenant clients match the shape production runs
+        run: python3 scripts/checks/check_realm_tenant_clients.py
+
       # h#727: static mode (no --live) is explicitly documented in the
       # script's own module docstring as "static, CI" — this wires that
       # documented, secrets-free half in. It cannot see holders (needs the
--- a/scripts/checks/check_checks_are_wired.py
+++ b/scripts/checks/check_checks_are_wired.py
@@ -100,7 +100,6 @@ TEST_SOURCES = ("bats", "pytest")
 # an acknowledgment of a known gap, not a decision that it is acceptable
 # forever.
 ALLOWLIST: dict[str, str] = {
-    "check_realm_tenant_clients.py": "test-only, found by h#736 — tracked in h#758",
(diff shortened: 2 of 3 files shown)


jonhill90:
Cannot auto-update — conflicts with `main`. The cause is structural rather than a mistake, and it will recur for all remaining PRs in this series: every one of the eight removes a line from the same `ALLOWLIST` in `check_checks_are_wired.py` and appends a step to the same region of `ci.yml`. #760 and #761 have both landed since this branch was cut, so the same two files have moved twice underneath it.

Rebase this on current `main` and re-push. Then, to stop paying this every time: cut each subsequent branch from `main` **after** the previous one merges, rather than cutting them all up front. Serial work on shared files wants serial branches.

File list checked first this time and it is clean — `ci.yml`, the allowlist, and the test, nothing else. I am doing that on every PR now: an unrelated 155-line file rode into #760 because I verified the wiring claim and never read the diff's file list. My miss, and exactly the failure this repo's own review rule exists to prevent.

jonhill90:
Verified before merging: on `main` this check appears zero times in `ci.yml` and once in the `ALLOWLIST`, so the gap is real and the fix does both halves in one PR — the wiring and the allowlist removal together, which is the rule #763 had to exist to enforce. File list is clean: `ci.yml`, the allowlist, and one test.

Thanks for taking the rebase rather than arguing for merging out of order. Three of eight.

PR #764, opened 2026-08-05
fix(checks): wire check_role_mappings_repointed.py into ci.yml (h#758 4/8)

jonhill90:
## h#758, check 4 of 8: `check_role_mappings_repointed.py`

Part of h#758 — the 8 checks h#736's fix found wired only to their own bats/pytest control. Each check gets its own investigation and its own PR.

**What it asserts.** No consumer's role-mapping configuration (Grafana's `role_attribute_path`, OpenBao's admin-sso `bound_claims`) names a legacy realm role — `admin`, `user`, `editor`, `viewer`. Those four are being removed because they have zero holders, so a mapping that still names one authorises nobody. Because Grafana's `role_attribute_strict` defaults to `false`, a half-repointed arm fails silently: no log line, no failed login. That is the exact shape #637/#666 produced — Stage 2a repointed Grafana's admin arm to `platform-admin` and left the editor arm naming a zero-holder role, and the config read as though the work was finished.

```
  ok    Grafana role_attribute_path: platform-admin, platform-editor
  ok    OpenBao admin-sso bound_claims: platform-admin

PASS — 2 mapping(s) checked, none names a legacy realm role
```

**Where it can run.** Fully hermetic — parses `docker-compose.observability.yml` and `scripts/vault.sh` statically, no live Keycloak needed. Same shape as checks 1-3/8 (#760, #761, #762): no design question about placement, just a check that had never been asked to run. Wired into `ci.yml` right after `check_retired_roles_ungranted.py`, its thematic sibling on role-scheme hygiene.

**Positive control, both directions.** Saved the `ci.yml` diff, reverted it, reran this check's full bats file — the new wiring-proof test failed on exactly the predicted assertion (invocation line absent from `ci.yml`), the 8 pre-existing logic tests stayed green. Reapplied the diff, reran — all 9 passed.

**`check_silent_success.py`.** Clean, no new finding (A=1, B=5, both the existing tracked baseline).

**Full suites, both green:**
```
bats tests/scripts/     591 passed, 0 failed
pytest tests/checks/     80 passed in 144.92s
```

**Bookkeeping.** Removes `check_role_mappings_repointed.py`'s own now-stale `ALLOWLIST` entry.

## Test plan
- [x] Wired into `ci.yml`
- [x] Full `bats tests/scripts/` — 591 passed, 0 failed
- [x] Full `pytest tests/checks/` — 80 passed
- [x] Stale `ALLOWLIST` entry removed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,19 @@ jobs:
       - name: Assert no retired role is granted anything (static)
         run: python3 scripts/checks/check_retired_roles_ungranted.py
 
+      # h#758 (found by h#736's fix): wired only to its own bats control
+      # since it was written. Fully hermetic — parses
+      # docker-compose.observability.yml and scripts/vault.sh statically for
+      # role-mapping expressions, no live Keycloak needed, no design question
+      # about WHERE it can run. Guards the exact half-repoint shape from
+      # #637/#666: Grafana's role_attribute_path had its admin arm repointed
+      # from the zero-holder legacy role to platform-admin, but its editor arm
+      # was left naming another zero-holder role — and because
+      # role_attribute_strict defaults to false, that arm fails silently, with
+      # no log line and no failed login to notice it by.
+      - name: Assert no consumer role mapping still names a legacy realm role
+        run: python3 scripts/checks/check_role_mappings_repointed.py
+
       - name: Assert no path revokes root before OIDC is enabled
         run: python3 scripts/checks/check_vault_revoke_order.py
 
--- a/scripts/checks/check_checks_are_wired.py
+++ b/scripts/checks/check_checks_are_wired.py
@@ -100,7 +100,6 @@ TEST_SOURCES = ("bats", "pytest")
 # an acknowledgment of a known gap, not a decision that it is acceptable
 # forever.
 ALLOWLIST: dict[str, str] = {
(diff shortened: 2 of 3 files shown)


jonhill90:
Holding this one deliberately until #762 lands, and flagging the sequencing rather than just rebasing around it.

There are now two open PRs in this series and both touch `.github/workflows/ci.yml` and `scripts/checks/check_checks_are_wired.py`. #762 is DIRTY and this one is BEHIND. If I merge this first, #762's conflict gets worse, not better — so the queue has to drain in order.

The structural fix is the one from #762's thread and it has not been applied yet: **cut each branch from `main` after the previous one merges**, rather than cutting several up front. Four of the eight are done or in flight; the remaining four should be strictly serial. Opening 5/8 before 3/8 has landed will keep costing rebases that have nothing to do with the work.

File list is clean on both — `ci.yml`, the allowlist, and one test each, nothing stray.

Order from here: rebase #762 → merge #762 → update this branch → merge this → then cut 5/8.

jonhill90:
Verified both halves against the files rather than the description: on `main`, zero occurrences in `ci.yml` and one in the `ALLOWLIST`; on the branch, one in `ci.yml` and zero in the allowlist. Wiring and suppression-removal in the same PR, as the standard now requires. File list clean — `ci.yml`, the allowlist, one control test.

Four of eight. The rebase cost was real and it is now behind us: the remaining four, each cut from `main` after the previous merges, should apply without any of this.

PR #767, opened 2026-08-05
fix(checks): wire realm-tenant-serves-test.sh into ci.yml (h#758 7/8)

jonhill90:
## h#758, check 7 of 8: `realm-tenant-serves-test.sh`

Part of h#758 — the 8 checks h#736's fix found wired only to their own bats/pytest control. Cut fresh from main after #766 (6/8) merged.

**What it asserts.** `platform-realm.json` actually SERVES the tenant, not just that the file is shaped correctly. It imports the real realm into a throwaway Keycloak container, assigns a client role, and reads the resulting access token — proving Keycloak actually emits `resource_access.hill90-ui.roles`, which depends on built-in client scopes this repo does not define. `check_realm_tenant_clients.py` (h#758 3/8, #762) shows the FILE is right; only a real Keycloak can show the CLAIM is right.

```
All 21 assertions passed against a real Keycloak.
  image   cached, no pull
  start   1s
  ready   25s  (8/60 attempts of a 120s cap)
  asserts 23s  (21 assertions, each one or more docker exec calls)
  total   49s accounted for
```

**Where it can run.** No design question, unlike `check_vault_covers_compose.py` (h#758 6/8, #766). This script builds and destroys its own throwaway Keycloak — it never touches a running stack, the VPS, or any live secret; the only credentials involved are literal throwaways it defines itself (see its own `argv-ok` comments). It needs only Docker, already used elsewhere in the same `ci.yml` job (promtool, `minio-readiness-test.sh` right above this step). Wired directly into `ci.yml` next to `minio-readiness-test.sh`.

Its only prior caller was `tests/checks/test_realm_tenant_serves.py`, a pytest wrapper — `pytest tests/checks/` running in CI is still not this SCRIPT being invoked for real, only the wrapper's own assertions about it, the same TEST-ONLY distinction as every other check in this series.

**Positive control, both directions.** Reverted `ci.yml`, the new wiring-proof pytest test failed on exactly the predicted assertion (invocation line absent), the other 2 tests in the file were unaffected. Reapplied, it passed.

**`check_silent_success.py`.** Clean, no new finding. Unlike #766, this step carries no `if:` condition at all, so shape B's scan doesn't even consider it — no `ALLOWLIST_B` entry needed.

**Full suites, both green:**
```
pytest tests/checks/     82 passed
```

**Bookkeeping, same PR as the wiring.** Removes `realm-tenant-serves-test.sh`'s own now-stale `ALLOWLIST` entry. Only `tenant-login-local-test.sh` (h#758 8/8) remains.

## Test plan
- [x] Established WHERE it can genuinely run — no design question this time, unlike 6/8
- [x] Wired directly into `ci.yml`
- [x] `check_silent_success.py` shows no new finding (no `if:` on the new step)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,6 +253,26 @@ jobs:
       - name: MinIO readiness regression test
         run: bash scripts/checks/minio-readiness-test.sh
 
+      # h#758 (found by h#736's fix): this script's only caller was
+      # tests/checks/test_realm_tenant_serves.py, a pytest wrapper — and
+      # `pytest tests/checks/` running in CI is still not this SCRIPT being
+      # invoked for real, only the wrapper's own assertions about it (see
+      # check_checks_are_wired.py's docstring on why TEST_SOURCES never
+      # counts). No design question about placement, unlike
+      # check_vault_covers_compose.py (h#758 6/8): this builds and destroys
+      # its OWN throwaway Keycloak container — it never touches a running
+      # stack, the VPS, or any live secret. The only credentials involved are
+      # literal throwaways the script defines itself (see its own argv-ok
+      # comments). It needs only Docker, already used elsewhere in this same
+      # job (promtool above, minio-readiness-test.sh right above this step).
+      # Proves what check_realm_tenant_clients.py (h#758 3/8) cannot: that
+      # file shows platform-realm.json is SHAPED correctly, this shows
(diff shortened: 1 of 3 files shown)


jonhill90:
Both halves verified against the files — zero/one on `main`, one/zero on the branch, wiring and suppression-removal in one PR. File list clean. Seven of eight.

PR #768, opened 2026-08-05
docs(checks): file tenant-login-local-test.sh as unreachable, not wired (h#758 8/8)

jonhill90:
## h#758, check 8 of 8: `tenant-login-local-test.sh` — FILED, not wired

Part of h#758 — the last of the 8 checks h#736's fix found wired only to their own bats/pytest control. This one does not end in a real caller, and that is the honest conclusion rather than a gap.

**What it asserts.** A completed authorization-code login against the LOCAL platform Keycloak, with a refusal control (a role-less user must be denied). Verified true right now — ran it directly against a locally running stack:
```
All 9 assertions passed against the local platform Keycloak.
```

**Where it can run — nowhere in this pipeline, and that is the answer.** Its own pytest wrapper's docstring already said so before this change: *"the only check in the suite that needs a RUNNING local platform stack, so it skips rather than fails when there isn't one — including in CI... Run it locally after `bash scripts/local.sh up`."* A step that skips reports green — wiring it into `ci.yml` would do exactly that on every PR, which is worse than leaving it unwired.

Checked against all three places this series asks a check to run before calling it unreachable:
- **CI** — no local platform stack exists on a GitHub-hosted runner. None of the other 7 checks needed one; `realm-tenant-serves-test.sh` (h#758 7/8, #767) proves the equivalent claim by building its own throwaway Keycloak instead of a full `local.sh` stack.
- **VPS deploy step** — `reusable-deploy-service.yml` deploys PRODUCTION; this script tests the LOCAL platform, a different Keycloak instance with different container names, hosts, and a separate local secrets store (`scripts/local.sh`'s own `$LOCAL_VAULT_DIR`, deliberately isolated from `infra/secrets/prod.enc.env`). There's no VPS concept this maps to.
- **Scheduled workflow** — none of the four existing ones (`audit-hill90-ui-client`, `deploy-drift`, `tenant-baseline-agreement`, `vault-sync-to-sops`) stand up an ephemeral local dev stack; all four check live production directly. Building one from scratch — full compose orchestration matching `compose/local.yml` plus overrides, hostname/DNS resolution, generating a fresh local age key and secrets — is new infrastructure this change does not add, and a decision for Jon to make deliberately, not one to improvise inside "wire the check."

**What this PR does.** Updates the `ALLOWLIST` reason in `check_checks_are_wired.py` from the placeholder `"tracked in h#758"` to a permanent, decided one, with the full reasoning above kept as a block comment so the next reader doesn't have to re-derive it. If that local-stack-in-CI infrastructure is ever built for another reason, this is the first check that should be pointed at it.

**Full suites, both green:**
```
pytest tests/checks/     82 passed
```

**`check_silent_success.py`.** Clean, no new finding — no workflow file touched.

**h#758 is now fully worked**: 7 of 8 checks wired for real (#760, #761, #762, #764, #765, #766, #767), 1 of 8 filed as genuinely unreachable rather than half-wired somewhere it would silently skip.

## Test plan
- [x] Check asserts something true about the estate right now (verified against a real local stack)
- [x] Established there is no place in this pipeline it can genuinely run — all three categories checked and ruled out
- [x] `ALLOWLIST` entry updated with the permanent decision, not the stale "tracked in h#758" placeholder
- [x] `check_silent_success.py` unaffected

### DIFF
--- a/scripts/checks/check_checks_are_wired.py
+++ b/scripts/checks/check_checks_are_wired.py
@@ -100,7 +100,57 @@ TEST_SOURCES = ("bats", "pytest")
 # an acknowledgment of a known gap, not a decision that it is acceptable
 # forever.
 ALLOWLIST: dict[str, str] = {
-    "tenant-login-local-test.sh": "test-only, found by h#736 — tracked in h#758",
+    # h#758 8/8, closed as FILED rather than wired — the only one of the eight
+    # that did not end in a real caller. It needs a RUNNING LOCAL PLATFORM
+    # STACK (docker inspect on a Keycloak container named from
+    # $CONTAINER_PREFIX, plus .env.local for host/port/secret config) — its
+    # own pytest wrapper's docstring already said so before this entry was
+    # written: "the only check in the suite that needs a RUNNING local
+    # platform stack, so it skips rather than fails when there isn't one —
+    # including in CI... Run it locally after `bash scripts/local.sh up`."
+    # Verified directly: the property IS true right now, run against a real
+    # local stack (all 9 assertions passed) — this is not a stale check, just
+    # one CI structurally cannot reach.
+    #
+    # Checked against all three places h#758 asks a check to run before
+    # calling it unreachable:
+    #   CI (ci.yml)      — no local platform stack exists on a GitHub-hosted
+    #                       runner, and none of the other 7 checks in this
+    #                       series needed one; realm-tenant-serves-test.sh
+    #                       (h#758 7/8) proves Keycloak emits the same claim
+    #                       without needing this stack, by building its own
+    #                       throwaway Keycloak instead of a full local.sh
+    #                       stack.
+    #   VPS deploy step   — reusable-deploy-service.yml deploys PRODUCTION;
+    #                       this script tests the LOCAL platform, a different
+    #                       Keycloak instance with different container names,
+    #                       hosts and a separate local secrets store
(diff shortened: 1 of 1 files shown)


jonhill90:
Verified the diff: a vague allowlist reason replaced with a documented one that names exactly what the check needs, cites its own pytest wrapper's docstring saying so, and records that the property was confirmed true against a real local stack. Correctly still allowlisted, because it genuinely is unwired — the entry now explains why instead of implying someone forgot.

**This is the right way to close the last of the eight, and refusing to wire it was the finding.** A step in `ci.yml` would have skipped on every PR and reported green, which is worse than no step. You checked all three places this series asks a check to run — CI, the deploy step, a scheduled workflow — and gave a specific reason each is wrong rather than a general shrug: no local stack on a hosted runner; the deploy workflow targets production while this tests a deliberately isolated local Keycloak with its own secrets store; and none of the four scheduled workflows stand up an ephemeral dev stack. Building one is new infrastructure and Jon's call, not something to improvise inside "wire the check."

That closes **h#758 and the whole h#736 finding**: 9 of 35 checks were wired only to their own tests. One was `check_checks_are_wired.py` itself, fixed in #759. Seven now run for real — #760, #761, #762, #764, #765, #766, #767. This one is filed as structurally unreachable with its reason recorded. Nine of nine resolved, and the two that mattered most were the check that would have caught the 2026-08-03 auth outage and the one that scans for the silent-success family, both of which had been sitting green and unrun.

PR #769, opened 2026-08-05
fix(checks): close two vacuous-pass gaps found by the sibling-drift sweep

jonhill90:
## Sibling-drift sweep: scripts/checks pairs

Companion to hill90-app's SSE/CLI silent-success sweep — same family, this repo's own `scripts/checks/` compared sibling-to-sibling instead of symptom-to-symptom.

**`check_vault_covers_compose.py`'s `declared()`** parses the exact same `vault_paths_for_service()` function, with the identical regex, as `check_declared_paths_are_seeded.py`'s `declared_paths()` — which was hardened after this exact drift happened for real once (h#730): `echo "..."` becoming `printf '%s' "..."` drops every case arm from the parse while the outer function-body regex still matches, so every real service reads as *"declares no vault paths -> SOPS path, which has all keys"* and is skipped before ever being compared. A run that checked zero services printed PASS. This check shared the exact same unguarded regex until now; `required_vars()` (`vault_required_vars_for_service()`) had the missing-function guard but not the empty-result one either.

**`check_root_revoke_fails_closed.py`** had the same shape one level up: `if not files: return 1` covers zero workflow *files*, but a `minting` list ending up empty (`MINT`'s invocation strings no longer matching anything real — a renamed or reshaped `vault.sh` subcommand) skipped the whole audit loop and printed *"PASS — every root-minting workflow revokes and verifies on failure"* having verified zero workflows. `vault-regain-root.yml` and `vault-init.yml` are known, current root-minting workflows, so an empty list means the check went blind, not that root minting stopped.

**Neither gap is currently live** — both checks report correctly against the real estate right now, verified directly before and after this fix. This closes the *resilience* gap: what happens the next time either regex-parsed function's shape drifts, the same way it already has once.

**Positive-controlled, both directions, both checks**: reused the exact h#730 mutation (echo → printf, dropping every case arm) for `check_vault_covers_compose.py`'s two functions, and a renamed-subcommand mutation for `check_root_revoke_fails_closed.py`'s `MINT` strings. Reverted each source fix, confirmed the new bats test fails for exactly the predicted reason, reapplied, confirmed green.

**Full suites, both green:**
```
pytest tests/checks/    82 passed in 154.86s
```

**Reporting-only**: neither fix changes what any deploy or workflow *does* — only what these two static checks report when their own parsing assumptions eventually drift.

## Test plan
- [x] Both checks still pass correctly against the real estate right now
- [x] Positive control: revert → new test fails for the predicted reason → restore → passes (both checks)
- [x] Full `bats tests/scripts/` — 596 passed, 0 failed
- [x] No production/deploy behavior touched

### DIFF
--- a/scripts/checks/check_root_revoke_fails_closed.py
+++ b/scripts/checks/check_root_revoke_fails_closed.py
@@ -212,6 +212,26 @@ def main() -> int:
     print(f"  workflows that mint a root token: {', '.join(minting) or 'none'}")
     print()
 
+    # `minting` ending up empty is indistinguishable from MINT's invocation
+    # strings no longer matching anything real — vault.sh's regain-root/init
+    # subcommands being renamed or reshaped — and would pass vacuously:
+    # nothing to audit, so `problems` stays empty and this prints PASS having
+    # verified zero workflows. vault-regain-root.yml and vault-init.yml are
+    # known, current root-minting workflows, so a genuinely empty list here
+    # means this check went blind, not that root minting stopped happening.
+    # Same shape as check_vault_covers_compose.py's declared()/required_vars()
+    # guards and check_declared_paths_are_seeded.py's h#730 fix.
+    if not minting:
+        print(
+            "FAIL — no root-minting workflow found at all. Either root minting has "
+            "genuinely been removed from every workflow (update MINT's comment and "
+            "this message if so, deliberately), or MINT no longer matches — a "
+            "renamed or reshaped vault.sh subcommand — which is a broken check, not "
+            "a clean estate.",
+            file=sys.stderr,
+        )
+        return 1
+
     if problems:
         print(f"FAIL — {len(problems)} problem(s):\n")
         for p in problems:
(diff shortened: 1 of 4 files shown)


jonhill90:
Merging. Comparing siblings rather than symptoms found the same 0==0 shape in two more checks, and reusing the **exact h#730 mutation** — `echo` becoming `printf`, dropping every case arm while the outer function-body regex still matches — as the positive control is the right instrument: it is not a synthetic mutation, it is the drift that already happened once, replayed against a check that shared the unguarded regex.

The `check_root_revoke_fails_closed.py` gap is the more interesting of the two because the guard existed at the wrong level: `if not files: return 1` catches zero workflow *files*, but an empty `minting` list — `MINT`'s invocation strings no longer matching a renamed `vault.sh` subcommand — skipped the audit loop entirely and printed *"every root-minting workflow revokes and verifies on failure"* having verified none. A guard one level up from where the count actually goes to zero looks like protection and is not.

Stating plainly that neither gap is live today, verified before and after, is the right framing: this closes a resilience gap against the next drift, not an active exposure. That distinction has mattered repeatedly today and it keeps the record honest.

PR #770, opened 2026-08-05
fix(workflows): scheduled-workflow guard drift — ConnectTimeout + preflight secrets

jonhill90:
## Sibling-drift sweep: the four scheduled workflows

Companion to hill90-app's silent-success sweep — compared the four scheduled GitHub Actions workflows against each other (structurally similar jobs, natural place for one to have a guard its siblings lack).

**ConnectTimeout.** 7 of `vault-sync-to-sops.yml`'s 9 SSH/SCP calls had no timeout at all, and the 8th used `ConnectTimeout=10` while every other call across all four scheduled workflows uses `15`. Doesn't change what a successful run does — only how long a genuinely unreachable VPS takes to fail loudly instead of hanging on OS-default TCP timeouts, which run far longer, especially for a silently-dropping-packets failure mode. Normalized all 9 calls to `15`.

**Preflight secrets guard.** `deploy-drift.yml` already refuses to run with a clear `::error::` when any of `TS_OAUTH_CLIENT_ID`/`TS_OAUTH_SECRET`/`VPS_SSH_PRIVATE_KEY`/`SOPS_AGE_KEY` is missing, rather than failing confusingly deep inside Tailscale/SSH auth. The other three used the identical four secrets for the identical purpose without it.

- `audit-hill90-ui-client.yml` and `vault-sync-to-sops.yml`: guard added up front, right after checkout — same placement as `deploy-drift.yml`, since both jobs' entire purpose depends on VPS/vault access from the start.
- `tenant-baseline-agreement.yml` is **not** symmetric with its three siblings, and the guard is placed differently on purpose: its actual comparison (`check_tenant_baseline_agrees.py`) needs none of the four secrets at all — only the heartbeat dead-man's-switch metric after it does, per that job's own existing comment. An unconditional preflight before the comparison would have blocked a check that doesn't need these secrets. Placed instead right before the heartbeat's `Setup Tailscale` step, gated `if: always()` to match every other step in that block, so a missing secret still fails loudly (distinct from a real baseline disagreement, which already failed the earlier step) without gating the comparison this job actually exists to run.

Pure early-exit in all three cases — never touches the VPS, so success paths are unaffected.

**`check_silent_success.py`**: clean, no new finding — none of the new step names match `CLEANUP_WORDS`.

**Full suites, both green:**
```
bats tests/scripts/    593 passed, 0 failed
```

## Test plan
- [x] All three files' YAML validated
- [x] No production/deploy behavior touched — pure reporting/timeout additions

### DIFF
--- a/.github/workflows/audit-hill90-ui-client.yml
+++ b/.github/workflows/audit-hill90-ui-client.yml
@@ -72,6 +72,30 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      # Same guard as deploy-drift.yml's — added by the sibling-drift sweep
+      # that found this workflow, tenant-baseline-agreement.yml and
+      # vault-sync-to-sops.yml using the identical four secrets for the
+      # identical purpose without it. Without this, a missing/blank secret
+      # surfaces later as a confusing Tailscale or SSH auth failure instead of
+      # a clear "cannot run" — the same CANNOT-DETERMINE-vs-clean-failure
+      # distinction this repo's own scripts/checks/*.py already take care
+      # over. Pure early-exit: never touches the VPS.
+      - name: Preflight — required secrets are present
+        env:
+          TS_OAUTH_CLIENT_ID: ${{ secrets.TS_OAUTH_CLIENT_ID }}
+          TS_OAUTH_SECRET: ${{ secrets.TS_OAUTH_SECRET }}
+          VPS_SSH_PRIVATE_KEY: ${{ secrets.VPS_SSH_PRIVATE_KEY }}
+          SOPS_AGE_KEY: ${{ secrets.SOPS_AGE_KEY }}
+        run: |
+          missing=()
+          for name in TS_OAUTH_CLIENT_ID TS_OAUTH_SECRET VPS_SSH_PRIVATE_KEY SOPS_AGE_KEY; do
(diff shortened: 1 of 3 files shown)


jonhill90:
Merging. Both halves are the shape I asked for, and the second is the better answer.

**ConnectTimeout**: 7 of 9 SSH/SCP calls in `vault-sync-to-sops.yml` had none, and the 8th used 10 where every other call across all four workflows uses 15. Nothing about a successful run changes — only how long a genuinely unreachable VPS takes to fail loudly rather than hang on OS-default TCP timeouts. The silently-dropping-packets case is where that default is worst, and a job that looks like it is working for minutes is its own version of today's family.

**The preflight guard is where you refused to assume symmetry, and were right.** Two siblings needed `deploy-drift.yml`'s guard in the same place. `tenant-baseline-agreement.yml` genuinely is not symmetric: `check_tenant_baseline_agrees.py` needs none of those four secrets — only the heartbeat step after it does. An unconditional preflight would have blocked the check that job exists to run, in order to protect a step that comes later. Placing it before `Setup Tailscale` with `if: always()`, matching that block's own convention, keeps a missing secret failing loudly, leaves the comparison ungated, and keeps it distinguishable from a real baseline disagreement that already fails earlier.

That is the instruction working as intended: sometimes the divergence is correct and the useful output is explaining why, not aligning it.

PR #772, opened 2026-08-05
docs: github-actions.md claimed deploy.yml ships the tenant's app services — it never has

jonhill90:
## Summary

Cross-repo doc/code drift sweep (paired with `hill90-app`#461). Three places in `docs/reference/github-actions.md` claimed the Deploy Workflow "Deploys application services (api, ai, mcp, auth, ui)" and auto-triggers on any push to `main`.

Verified against `deploy.yml` directly: its `workflow_dispatch` options and push-trigger `paths:` are `[all, db, auth, minio, vault, observability]` — platform stacks only. `api`, `ai`, `mcp` and `ui` never appear in `deploy.yml` or any `deploy-*.yml` anywhere in this repo. Those are `hill90-app`'s own services, deployed entirely by that repo's own separate pipeline — per this repo's own `CLAUDE.md` and `docs/runbooks/tenant-app-deployment.md`.

This file's own sibling, `docs/reference/deployment.md`, states the correct architecture ("Three stacks... Edge, Vault, Observability") and was rewritten in the 2026-07-29 correction sweep ("correct five statements the tenant deploy made false") that evidently never reached this file — they sit in the same directory making opposite claims about the same workflow.

## Why this ranks high

A reader who opens this file — a plausible first stop, cross-linked from `CONTRIBUTING.md`'s reference map — would believe pushing to `main` here deploys the entire application stack including `api`/`ai`/`mcp`/`ui` with production certificates, and would look for the tenant's deploy path in the wrong repo and wrong workflow entirely. That's a claim someone would act on, not a stale count.

## What changed

Fixed all three occurrences (the workflow's own section, the numbered "Available GitHub Actions Workflows" list, and the step-by-step recreate walkthrough) to state what `deploy.yml` actually deploys, cross-reference `deployment.md`'s accurate framing, and point at `docs/runbooks/tenant-app-deployment.md` for where the app's own deploy actually lives.

Doc-only change; no code/workflow behavior touched.

### DIFF
--- a/docs/reference/github-actions.md
+++ b/docs/reference/github-actions.md
@@ -140,8 +140,11 @@ Policy Gate (Advisory) uses `gate-${{ github.head_ref }}` with `cancel-in-progre
 **Status:** Operational (consolidated single workflow)
 
 **Triggers:**
-- Push to `main` branch (auto)
-- Manual workflow dispatch
+- Push to `main` branch, scoped to specific platform paths (auth, db, minio,
+  vault, observability compose/config files — see `deploy.yml`'s own
+  `paths:` list)
+- Manual workflow dispatch, `service` input: `all | db | auth | minio |
+  vault | observability`
 
 **Certificates:**
 - Let's Encrypt PRODUCTION environment by default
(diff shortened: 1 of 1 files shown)


jonhill90:
Verified against `main` rather than the PR body: `deploy.yml`'s service input offers `[all, db, auth, minio, vault, observability]` — `api`, `ai`, `mcp` and `ui` appear nowhere. The doc asserted a capability the workflow has never had. Merging.

This is the finding from the code-vs-comment sweep that a reader would actually have **acted** on, which was that seam's ranking rule. Someone trusting it would hunt the tenant app's deploy path in this repo and this workflow — neither of which has it — and could believe a push to `main` ships app services when it ships none of them. Worse than a stale sentence: it sends the reader confidently to the wrong place. And this repo's own `docs/reference/deployment.md` already states the true three-stack architecture, so the correct information was one file away the whole time.

Same shape as CLAUDE.md invariant 2 earlier today — a document describing a capability with more confidence than the code ever justified — and the same right fix: correct the claim and state what is actually true, rather than deleting the sentence and leaving a gap where a reader expects an answer.

PR #773, opened 2026-08-05
fix(scripts): docker network create's exit status, explicitly checked

jonhill90:
## Sibling-drift sweep: scripts/deploy.sh vs scripts/local.sh

Both scripts ran \`docker network create\` as a bare statement with no exit-status check, printing a hardcoded success line unconditionally afterward.

**Corrected from the initial sweep report for \`deploy.sh\` specifically**, after verifying rather than assuming: this script has \`set -e\` at the top, and a forced-failure test confirms it already aborts the script the instant \`docker network create\` fails — the echo never runs, docker's own exit code propagates. So this was **not** a live silent-success bug in \`deploy.sh\`; it relied entirely on that interpreter flag staying set, with docker's own raw error text standing in for a legible message. Made explicit anyway (\`|| die \"...\"\`), so it doesn't depend on an interpreter flag surviving three future refactors, and to match \`local.sh\`'s fix with the same shape rather than leaving one silent and one explicit for the identical operation.

\`local.sh\` has **no** \`set -e\` (only \`set -uo pipefail\`), and this loop's bare \`docker network create\` genuinely printed \`"Created \${netpfx}_\${net}"\` unconditionally regardless of whether the create succeeded. Fixed with the same explicit \`|| die\` \`local.sh\`'s own \`cmd_sso\`/\`cmd_vault\` already use for every other risky call in this script.

**Positive-controlled**: a forced-failure harness (\`docker()\` stubbed to fail on \`network create\`) confirms both fixed scripts now stop with a clear \"Failed to create ... network\" message and non-zero exit, and confirms the **unfixed** \`local.sh\` code reaches its end with exit 0 having silently reported success on the identical simulated failure.

Not retrofitting \`set -e\` onto all of \`local.sh\` here — that's a separate, larger decision, and this fix doesn't depend on it.

**Full suites, both green:**
\`\`\`
\`\`\`

\`shellcheck --severity=error\`: clean on both files.

## Test plan
- [x] Verified empirically whether \`set -e\` already caught this in \`deploy.sh\` — it does; framing corrected accordingly rather than reporting a false positive
- [x] Positive control: forced-failure harness confirms both fixes stop cleanly; confirms pre-fix \`local.sh\` silently "succeeds"
- [x] Full \`bats tests/scripts/\` — 596 passed, 0 failed
- [x] Full \`pytest tests/checks/\` — 82 passed
- [x] \`shellcheck --severity=error\` clean

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -277,12 +277,24 @@ cmd_infra() {
 
     # Create internal networks if not present (edge compose creates hill90_edge;
     # internal networks are needed by app services but not by edge services)
+    #
+    # Sibling-drift sweep: this repo's own set -e already aborts the script if
+    # docker network create fails here — verified directly, not assumed: a
+    # forced failure never reaches the echo below and preserves docker's own
+    # exit code. So this was never a silent-success bug in practice, but it
+    # relied on that entirely on an implicit bash abort with docker's own raw
+    # error text, rather than a legible message naming what failed — and the
+    # identical code in local.sh has no such backstop (see that fix). Made
+    # explicit here too, for the same reason a check should not depend on an
+    # interpreter flag staying set three refactors from now.
     if ! docker network inspect hill90_internal >/dev/null 2>&1; then
-        docker network create --driver bridge --internal hill90_internal
+        docker network create --driver bridge --internal hill90_internal \
+            || die "Failed to create hill90_internal network"
         echo "✓ Created hill90_internal network for app services"
     fi
     if ! docker network inspect hill90_agent_internal >/dev/null 2>&1; then
-        docker network create --driver bridge --internal hill90_agent_internal
+        docker network create --driver bridge --internal hill90_agent_internal \
+            || die "Failed to create hill90_agent_internal network"
         echo "✓ Created hill90_agent_internal network for agent containers"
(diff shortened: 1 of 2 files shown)


jonhill90:
Merging — and the self-correction is the most valuable thing in this PR.

Your initial sweep reported this as a live silent-success in **both** scripts, and I relayed that upward as "the same family found a fifth and sixth time." You then verified rather than assumed: `deploy.sh` has `set -e`, so it already aborts the instant `docker network create` fails and the success line never prints. I confirmed that independently against `main` — `deploy.sh` carries `set -e`, `local.sh` carries only `set -uo pipefail`. One live instance, not two, and I have corrected the count I gave.

Fixing both anyway is right for the reason you gave rather than for symmetry: `deploy.sh`'s safety rested entirely on an interpreter flag surviving future refactors, with docker's raw error text standing in for a legible message. Making it explicit costs nothing and removes a dependency nobody would notice breaking.

The positive control is the right shape — stubbing `docker` to fail, showing the **unfixed** `local.sh` reaching its end at exit 0 having printed "Created …", and the fixed version stopping with a clear message. Bug and fix demonstrated on the same harness, rather than one asserted and the other tested.

PR #774, opened 2026-08-05
fix(compose): extend minio.yml's ${VAR:?} required-secret guard to 3 more files

jonhill90:
## Sibling-drift sweep: the six prod compose files

5 of 6 prod compose files used bare \`\${VAR}\` for equally sensitive values with no guard, while \`minio.yml\` alone used \`\${VAR:?...}\` — deliberately, with a comment explaining why: an unset MinIO root password silently starts the server on \`minioadmin/minioadmin\` with only a WARN. The estate has already paid for the opposite of this pattern twice: the 2026-08-03 outage was \`GRAFANA_OIDC_CLIENT_SECRET\` resolving to empty and Grafana deploying anyway, and #766 (h#758 6/8) wired a whole check (\`check_vault_covers_compose.py\`) specifically to catch this class after the fact. A deploy that refuses to start with a missing secret is strictly better than one that starts broken.

**Extended to 7 variables across 2 files that needed it — not 5 files.** \`infra.yml\` and \`vault.yml\` turned out to have no bare-\`\${VAR}\` secrets at all (every one already carries an explicit \`:-default\`), so they're untouched.

- \`auth.yml\`: \`DB_USER\`, \`DB_PASSWORD\`, \`KC_ADMIN_USERNAME\`, \`KC_ADMIN_PASSWORD\`, \`VAULT_OIDC_CLIENT_SECRET\`
- \`db.yml\`: \`DB_USER\`, \`DB_PASSWORD\` (all 5 occurrences each, including the healthcheck)
- \`observability.yml\`: \`GRAFANA_ADMIN_PASSWORD\`, \`GRAFANA_OIDC_CLIENT_SECRET\`

**Deliberately excluded: \`auth.yml\`'s \`HILL90_UI_CLIENT_SECRET\`**, found during the same sweep and NOT guarded — verified it is not SOPS/vault-backed at all, and \`keycloak.sh\`'s own handling already refuses on empty *only* when the \`hill90-ui\` client doesn't yet exist ("Matters only on a FIRST import... \`start --import-realm\` ignores an existing realm" — that file's own comment). It's legitimately empty on every routine deploy after the first. Guarding it at the compose level would have broken every routine auth deploy — exactly the production-behavior risk this task was told to avoid, caught by the verification step before it shipped rather than after.

**Verified against the current SOPS surface before landing**, per instruction, so this doesn't arm a guard that fires on the next real deploy: all 7 guarded variables confirmed present with non-trivial values in \`infra/secrets/prod.enc.env\`, independently corroborated by \`check_vault_covers_compose.py\`'s own "needs N, vault carries N" pass for auth/db/observability (unchanged before and after this fix).

**Positive-controlled, both directions, all three files**: \`docker compose config\` with all placeholders succeeds; removing any one placeholder makes the affected file refuse with the exact declared message.

**Three call sites needed updating** to keep working — all found by actually running the full suite, not assumed clean:
- \`ci.yml\`'s own "Validate compose files" step only supplied MinIO's three placeholders; added the new 7.
- \`check_volume_names.py\` maintains its own \`REQUIRED_PLACEHOLDERS\` list for the same reason \`ci.yml\` does; same fix.
- Two \`local.bats\` tests asserted "resolves to production names with NO environment set" against \`observability.yml\` directly; updated to supply just the two required placeholders while leaving every other var genuinely unset. One \`platform-services.bats\` test grepped an exact bare-form string the new guard reformats; loosened to a regex tolerant of any \`:?\`/\`:-\` suffix — the same "don't grep a literal a legitimate change can reformat" lesson already applied once today in the sibling \`hill90-app\` repo.
- \`scripts/validate.sh\` needed **no** change: it already detects a "required variable" failure and retries with real secrets via \`sops exec-env\`, explicitly designed around \`minio.yml\`'s existing guard as the precedent. Confirmed working unmodified: all four now-guarded files report \`✓ (needed secrets)\`.

\`check_vault_covers_compose.py\`, \`check_env_surface.py\`, \`check_secrets_schema.py\`, \`check_silent_success.py\`, \`check_checks_are_wired.py\`: all still pass, unchanged output.

**Full suites, both green:**
\`\`\`
\`\`\`

\`shellcheck --severity=error\`: clean.

## Test plan
- [x] Verified every guarded var populated in the current SOPS store before landing
- [x] Excluded a var (\`HILL90_UI_CLIENT_SECRET\`) after verification showed guarding it would break routine deploys
- [x] Positive control: all placeholders present → succeeds; any one missing → refuses with exact message
- [x] All three broken call sites found and fixed (\`ci.yml\`, \`check_volume_names.py\`, 2 bats tests + 1 grep)
- [x] \`scripts/validate.sh\` confirmed working unmodified against all newly-guarded files
- [x] Full \`bats tests/scripts/\` — 596 passed, 0 failed
- [x] Full \`pytest tests/checks/\` — 82 passed
- [x] \`shellcheck --severity=error\` clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,10 +44,26 @@ jobs:
         # an empty MinIO root password silently starts the server on
         # minioadmin/minioadmin with only a WARN. Supply throwaway values here —
         # this step validates STRUCTURE, not secrets.
+        #
+        # Sibling-drift sweep: minio.yml was the only prod compose file using
+        # this pattern; auth.yml, db.yml and observability.yml had the
+        # identical bare-${VAR} shape for equally sensitive values (Keycloak's
+        # own admin bootstrap credential among them) with no guard at all.
+        # Extended to match — see those files' own :? messages for the full
+        # list. Every var below is verified populated in the current SOPS
+        # store before this landed, so this only changes what happens on a
+        # FUTURE missing secret, not today's deploy.
         env:
           MINIO_ROOT_USER: placeholder
           MINIO_ROOT_PASSWORD: placeholder
           MINIO_OIDC_CLIENT_SECRET: placeholder
+          DB_USER: placeholder
+          DB_PASSWORD: placeholder
+          KC_ADMIN_USERNAME: placeholder
+          KC_ADMIN_PASSWORD: placeholder
+          VAULT_OIDC_CLIENT_SECRET: placeholder
+          GRAFANA_ADMIN_PASSWORD: placeholder
+          GRAFANA_OIDC_CLIENT_SECRET: placeholder
         run: |
           for f in deploy/compose/prod/docker-compose*.yml; do
             echo "Validating $f..."
--- a/deploy/compose/prod/docker-compose.auth.yml
+++ b/deploy/compose/prod/docker-compose.auth.yml
@@ -27,8 +27,8 @@ services:
     environment:
       - KC_DB=postgres
       - KC_DB_URL=jdbc:postgresql://postgres:5432/keycloak
-      - KC_DB_USERNAME=${DB_USER}
-      - KC_DB_PASSWORD=${DB_PASSWORD}
+      - KC_DB_USERNAME=${DB_USER:?DB_USER must be set}
+      - KC_DB_PASSWORD=${DB_PASSWORD:?DB_PASSWORD must be set}
       - KC_HOSTNAME=${KC_HOSTNAME:-https://auth.hill90.com}
       - KC_PROXY_HEADERS=xforwarded
       - KC_HTTP_ENABLED=true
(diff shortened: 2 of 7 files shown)


jonhill90:
Merging. The verification I required before this could ship paid for itself twice over.

It narrowed the scope honestly: five files became **two that needed it**, because `infra.yml` and `vault.yml` already carry explicit `:-default` on every variable. A hit is not a finding, applied to your own initial count.

And it caught the one that must **not** be guarded. `HILL90_UI_CLIENT_SECRET` is legitimately empty on every routine auth deploy after the first import — `start --import-realm` ignores an existing realm, as `keycloak.sh`'s own comment says — so a compose-level guard would have broken every auth deploy from here on. That is precisely the production-behaviour risk the hold existed to prevent, found by the verification step rather than by a failed deploy. Reporting it as a deliberate exclusion with the reason, instead of quietly omitting it, is what makes the exclusion reviewable.

All seven guarded variables confirmed present with non-trivial values in the current SOPS store, independently corroborated by `check_vault_covers_compose.py`'s own needs-N/carries-N pass for the three services — two methods, not one.

The rationale stands on its own: the 2026-08-03 outage was `GRAFANA_OIDC_CLIENT_SECRET` resolving to empty and Grafana deploying anyway, and #766 wired an entire check to catch that class after the fact. A deploy that refuses to start beats one that starts broken.

This touches three deploy-filtered compose files, so merging deploys auth, db and observability. Watching it through and re-verifying the baseline after.

PR #779, opened 2026-08-05
fix(deploy): make a manual deploy target the service that was asked for

jonhill90:
## Plan

`gh workflow run deploy.yml -f service=auth` deployed **all five** services. The `service` input has never selected anything.

`deploy.yml` OR-s two sources per job: `needs.changes.outputs.<svc> == 'true'`, and `github.event.inputs.service`. On a `workflow_dispatch` the first source is not merely unhelpful, it is actively wrong:

- the dispatch payload has no `before` field, so `dorny/paths-filter` warns and falls back to *"changes will be detected from last commit"*;
- `actions/checkout` defaults to `fetch-depth: 1`, so that commit has no parent;
- `git log --format= --no-renames --name-status -z -n 1` against a parentless commit lists the **entire repository** as `[added]`.

From run `31040415190`'s own log:

```
##[warning]'before' field is missing in event payload - changes will be detected from last commit
Detected 272 changed files
##[group]Filter db = true
  deploy/compose/prod/docker-compose.db.yml [added]
##[group]Filter auth = true
  deploy/compose/prod/docker-compose.auth.yml [added]
...
```

**Deepening the checkout does not fix this.** The fallback is triggered by the missing payload field, not by clone depth — `fetch-depth: 0` would still take the last-commit path.

The fix gates change detection to non-dispatch events. A skipped step leaves its outputs empty, so every `== 'true'` arm is false and only the explicit `inputs.service` arms decide. **Push behaviour is byte-for-byte unchanged.**

## Why this surfaced now, and what it cost

Not as a red job in the service anyone was looking at.

h#774 merged and its deploy failed on a git race between concurrent jobs (#778), leaving `db.yml` and `auth.yml` undeployed while `observability.yml` went live — a partial deploy reported as one failed job. Recovering it service by service is what exposed this: `-f service=auth` fanned out to all five, and the Keycloak restart landed underneath the concurrently running observability job's Grafana login check, which failed with `no Keycloak login form in the response`.

That failure was **transient, and verified so rather than assumed** — the same authorize URL serves the form now, fetched by hand rather than by re-running the check:

```
$ curl -o /dev/null -w '%{redirect_url}' https://grafana.hill90.com/login/generic_oauth
https://auth.hill90.com/realms/platform/protocol/openid-connect/auth?...
$ curl "$that" | python3 -c '...re.search(r"<form[^>]*id=\"kc-form-login\"")...'
LOGIN FORM PRESENT
```

So the report named the wrong service, and the operator who asked for one service got the estate. Both of today's deploy failures reduce to *too many jobs touching production at once*; #778 is the other half and is **not** fixed here.

## Risks

- **Genuinely narrow.** One `if:` on one step, plus a check and its control. No job's logic changes; on `push` the filter runs exactly as before.
- **The real risk is the reverse of the defect**: if the gate were inverted or pointed at the wrong event, a dispatch would deploy *nothing* and say nothing. The control requires red on both of those specific mutations, not merely on the absence of a gate.
- This does not touch #778. Until that is fixed, a **push** touching several filtered paths still fans out concurrently and can still race on `/opt/hill90/app`.

## Rollback

Revert the commit. Behaviour returns to "any dispatch deploys everything" — noisy and over-broad, never silent.

## Validation Evidence

Positive control, every arm requiring red, run locally:

```
$ bats tests/scripts/dispatch-targets-one-service-control.bats
1..6
ok 1 CONTROL: passes on the real tree
ok 2 the pre-fix step — no gate at all — is caught
ok 3 an inverted gate is caught, not accepted as 'mentions workflow_dispatch'
ok 4 a gate on some other event is caught
ok 5 a missing paths-filter step is CANNOT DETERMINE, not a pass
ok 6 ci.yml genuinely invokes this check, not just mentions it
```

Test 2 replays the pre-fix step **verbatim** — strips the `if:`, requires exit 1 — so the control is anchored to the defect that actually happened rather than to a rule written afterwards. Test 5 is the 0-inputs-scanned guard: renaming the step away must report blindness, not a clean run over nothing found. Test 6 is the h#736/h#758 wiring assertion.

Check on the real tree, and both workflows still parse:

```
$ python3 scripts/checks/check_dispatch_targets_one_service.py
PASS — change detection is gated to non-dispatch events, so a manual
       deploy is decided by the `service` input alone.

$ python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/deploy.yml')); print(', '.join(d['jobs']))"
changes, deploy-db, deploy-auth, deploy-minio, deploy-vault, deploy-observability
```

**What this evidence does NOT cover, stated rather than left implied.** The check is static — it reads `deploy.yml`. It cannot simulate a dispatch, because that needs GitHub. The runtime proof is a real single-service dispatch after merge, showing only that service's job running. I will run one and post the job list here. Until that comment exists, treat the runtime behaviour as unproven.

## Estate state throughout

Checked after each of today's deploys; degraded at no point:

```
missing: none  unhealthy:0  restarting:0  tenant:7
  hill90.com 200
  auth.hill90.com/realms/platform 200
```

All five services are now deployed at h#774's HEAD (`a25974e`), which the partial deploy had left half-applied.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,12 @@ jobs:
       - name: Assert CLAUDE.md's alert counts match the rules that ship
         run: python3 scripts/checks/check_alert_counts_documented.py
 
+      # `-f service=auth` deployed all five services, because on a
+      # workflow_dispatch paths-filter falls back to a parentless last commit
+      # and reports the whole repo as changed. Hermetic — reads deploy.yml.
+      - name: Assert a manual deploy targets one service
+        run: python3 scripts/checks/check_dispatch_targets_one_service.py
+
       - name: Install bats
         run: |
           sudo apt-get update
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,8 +40,22 @@ jobs:
       observability: ${{ steps.filter.outputs.observability }}
     steps:
       - uses: actions/checkout@v4
+      # Change detection is for `push` only. On a workflow_dispatch the event
+      # payload carries no `before`, so paths-filter falls back to "changes in
+      # the last commit" — and actions/checkout's default `fetch-depth: 1`
+      # leaves that commit parentless, so `git log --name-status -n 1` lists
+      # the ENTIRE repository as added. Run 31040415190 detected "272 changed
+      # files" and set all five filters true, which is why `-f service=auth`
+      # deployed all five services and restarted Keycloak underneath the
+      # observability job's login check.
+      #
+      # Skipping the step leaves every output empty, so the `== 'true'` arms
+      # below are all false and only the explicit `inputs.service` arms decide.
+      # Deepening the checkout would NOT fix this: the fallback is triggered by
+      # the missing `before` field, not by the clone depth.
       - uses: dorny/paths-filter@v3
         id: filter
+        if: github.event_name != 'workflow_dispatch'
         with:
           filters: |
             db:
--- a/scripts/checks/check_dispatch_targets_one_service.py
+++ b/scripts/checks/check_dispatch_targets_one_service.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+"""A manual deploy must deploy the service that was asked for, and no other.
+
+THE DEFECT (run 31040415190, 2026-08-05). `deploy.yml` decides what to deploy
+from two independent sources OR'd together: dorny/paths-filter output, and the
+`service` workflow_dispatch input. On a dispatch the first source is not merely
+unhelpful, it is actively wrong — the event payload has no `before` field, so
+paths-filter falls back to "changes in the last commit", and actions/checkout's
+default `fetch-depth: 1` leaves that commit with no parent. `git log
+--name-status -n 1` against a parentless commit reports the whole repository as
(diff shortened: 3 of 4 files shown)


jonhill90:
**Runtime proof, as promised in the PR body — the part the static check could not cover.**

Dispatched a single service against `main` after the merge:

```
$ gh workflow run deploy.yml -f service=vault --ref main
```

Run `31041549055`, jobs:

```
changes                        success
deploy-vault / Deploy vault    success
deploy-observability           skipped
deploy-db                      skipped
deploy-auth                    skipped
deploy-minio                   skipped
```

One service asked for, one service deployed, four skipped. Compare run `31040415190` before the fix, where `-f service=auth` ran all five.

Note the `changes` job still shows `success` — the job runs, its paths-filter *step* is skipped, and the empty outputs make every `== 'true'` arm false. That is the intended shape, and worth stating because a green `changes` job is exactly what someone would misread as change detection still deciding things.

Estate verified after, unchanged:

```
missing: none  unhealthy:0  restarting:0  tenant:7
  hill90.com 200
  auth.hill90.com/realms/platform 200
```

Still open and untouched by this: #778 — concurrent deploy jobs on a **push** touching several filtered paths, which race both on `/opt/hill90/app` and on the running estate.

PR #780, opened 2026-08-05
fix(vault): host-checkout cleanup could discard a secret its own copy-back failed to capture

jonhill90:
## Summary

Error-path sweep — code that only runs when something else has already failed, and structurally under-tested for exactly that reason.

`vault-init.yml`, `vault-reinitialize.yml`, and `vault-sync-to-sops.yml` each run `scripts/vault.sh store-unseal-key`/`sync-to-sops` on the VPS, which writes fresh secret material (a new unseal key, a sync token, synced values) into the host's **local, uncommitted** copy of `infra/secrets/prod.enc.env`. The next step `scp`s that file back to the runner so it can be committed — the whole point, per `vault-init.yml`'s own comment, being *"the key must survive loss of the host"*. The step after **that** reverts the host's local checkout with `git checkout -- infra/secrets/prod.enc.env`, so the host doesn't sit there permanently modified between runs.

That revert step was `if: always()`, unconditionally — including when the copy-back that was supposed to capture the new secret material never succeeded (a network blip mid-`scp`, or the post-copy decrypt/length check failing). In that case the host's copy — the only place the freshly-written secret existed before the copy-back — got discarded before anyone had a durable copy of it. Not silent (the copy-back step's own failure stays visible in the job log), but it destroyed **recoverable state the run existed to produce**, during exactly the disaster-recovery scenarios these workflows are for.

## Fix

Each revert step's `if:` now references its copy-back step's outcome (`always() && steps.copy-back.outcome != 'failure'`), so a failed copy-back leaves the host's mutated state in place for a retry or manual recovery instead of discarding it. `!= 'failure'` rather than `== 'success'` because `vault-sync-to-sops.yml`'s copy-back is itself conditional (skipped when there's no drift to sync) and the host should still be cleaned in that case.

## New durable check

`scripts/checks/check_vault_cleanup_guard.py`, matching this repo's own established static-check + bats-positive-control convention (see `check_vault_revoke_order.py`): parses each workflow for a `git checkout -- infra/secrets/prod.enc.env` revert step, walks back to its nearest preceding copy-back (`scp`) step, and fails unless the revert's `if:` references that step's id/outcome. Refuses to pass vacuously if it finds zero revert steps anywhere. Wired into `ci.yml`.

`tests/scripts/vault-cleanup-guard-control.bats` mutates copies to confirm the check goes red for each of five distinct reasons (missing id, reverted to bare `always()`, missing `if:` entirely, no preceding copy-back, zero workflows) plus the vacuous-pass guard, alongside a CONTROL test that the real repo passes.

## Test plan

- [x] `check_vault_cleanup_guard.py` passes on the fixed workflows
- [x] `check_vault_revoke_order.py` (unrelated, regression check) still passes
- [x] All 7 new bats tests pass
- [x] Full `tests/scripts/` suite (602 tests) — zero regressions
- [x] All four touched workflow YAML files parse

No infra/secrets/sops/credential work — this changes workflow trigger conditions and adds a static check, not secret values or vault behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,6 +199,15 @@ jobs:
       - name: Assert no path revokes root before OIDC is enabled
         run: python3 scripts/checks/check_vault_revoke_order.py
 
+      # Found in an error-path sweep: three workflows' "revert the host's
+      # SOPS checkout" cleanup step ran unconditionally (always()), including
+      # when the copy-back it depends on had already failed — discarding the
+      # only copy of freshly-written secret material before it was ever
+      # captured off-host. Wired only to its own bats control since it was
+      # written; fully hermetic, no vault, no host.
+      - name: Assert vault workflow cleanup steps don't discard un-captured secrets
+        run: python3 scripts/checks/check_vault_cleanup_guard.py
+
       # h#758 (found by h#736's fix): wired only to its own bats control
       # since it was written. Fully hermetic — parses scripts/_common.sh and
       # scripts/vault.sh statically, no vault, no host. Same shape as
--- a/.github/workflows/vault-init.yml
+++ b/.github/workflows/vault-init.yml
@@ -164,6 +164,7 @@ jobs:
             'cd /opt/hill90/app && export SOPS_AGE_KEY_FILE=/opt/hill90/secrets/keys/keys.txt && bash scripts/vault.sh store-unseal-key'
 
       - name: Copy updated SOPS back
+        id: copy-back
         run: |
           scp -i ~/.ssh/vps_key -o StrictHostKeyChecking=no \
             deploy@${{ steps.get-ip.outputs.tailscale_ip }}:/opt/hill90/app/infra/secrets/prod.enc.env \
@@ -173,8 +174,16 @@ jobs:
           echo "OPENBAO_UNSEAL_KEY decrypts to $LEN bytes"
           [ "$LEN" -gt 0 ] || { echo "::error::OPENBAO_UNSEAL_KEY missing after copy-back"; exit 1; }
 
+      # Gated on copy-back having actually captured the new unseal key, not a
+      # bare always(). This step only exists to clean the host's local
+      # checkout NOW THAT the key is safely in git — if copy-back failed
+      # (network blip mid-scp, or the decrypt check above), the host's copy
+      # is the only place that key existed. Reverting it anyway would discard
+      # the one thing this whole run existed to produce, during exactly the
+      # disaster-recovery scenario this workflow is for. See
+      # scripts/checks/check_vault_cleanup_guard.py.
       - name: Restore host checkout
-        if: always()
+        if: always() && steps.copy-back.outcome != 'failure'
         run: |
           ssh -i ~/.ssh/vps_key -o StrictHostKeyChecking=no \
             deploy@${{ steps.get-ip.outputs.tailscale_ip }} \
--- a/.github/workflows/vault-init.yml
+++ b/.github/workflows/vault-init.yml
@@ -174,16 +174,25 @@ jobs:
           echo "OPENBAO_UNSEAL_KEY decrypts to $LEN bytes"
           [ "$LEN" -gt 0 ] || { echo "::error::OPENBAO_UNSEAL_KEY missing after copy-back"; exit 1; }
 
-      # Gated on copy-back having actually captured the new unseal key, not a
-      # bare always(). This step only exists to clean the host's local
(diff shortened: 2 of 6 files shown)


jonhill90:
Reviewed against the workflows and `scripts/vault.sh`, not the PR body. **The fix is right and I want it in** — one real residual hole, and one claim that is overstated for the workflow it is written in.

## The hole: `outcome != 'failure'` also passes when copy-back never ran

In `vault-init.yml` and `vault-reinitialize.yml` the copy-back step carries no `if:`, so it can fail to run only because an *earlier* step failed. When that happens its outcome is `skipped` (or empty), never `failure` — so the guard passes and `Restore host checkout` still reverts.

That is reachable, and `cmd_store_unseal_key` is the proof. It writes first and verifies second:

```sh
if sops --set "[\"OPENBAO_UNSEAL_KEY\"] ${escaped}" "$SECRETS_FILE"; then
    success "✓ OPENBAO_UNSEAL_KEY written to $(basename "$SECRETS_FILE")"
...
stored_len=$(sops -d --extract '["OPENBAO_UNSEAL_KEY"]' "$SECRETS_FILE" ... )
[ "$stored_len" -gt 0 ] || die "OPENBAO_UNSEAL_KEY did not round-trip after writing."
```

So there is a window where the file **has been written** and the step **exits non-zero**. The round-trip check exists precisely because that write can produce something unreadable — this is the author's own anticipated failure, not a hypothetical. In that window: store-unseal-key = failure, copy-back = skipped, and your guard lets the revert through. The defect this PR exists to close survives one step upstream.

Suggested gate, which also reads more plainly than the negation:

- `vault-init.yml`, `vault-reinitialize.yml` — copy-back has no `if:`, so `skipped` is never legitimate there: `if: always() && steps.copy-back.outcome == 'success'`
- `vault-sync-to-sops.yml` — copy-back is genuinely conditional, so distinguish "skipped because there was no drift" from "skipped because something upstream died": `if: always() && (steps.compare.outputs.changed != 'true' || steps.copy-back.outcome == 'success')`

Your check already accepts `== 'success'`, so only the workflows change. Please add a control arm for it — a revert gated on `!= 'failure'` where copy-back was **skipped** must go red, and today it would not.

## The overstated claim

The comment in `vault-init.yml` says the host's copy "is the only place that key existed". Not for that workflow. `cmd_store_unseal_key` *reads* the key from `UNSEAL_KEY_PATH` — `/opt/hill90/secrets/openbao-unseal.key` — which the revert does not touch, so after a bad revert the key is still on the host and re-runnable. The same goes for `sync-to-sops`, whose values come from vault and can be re-synced.

That does not make this not-a-bug: a revert that silently discards freshly written secret material during disaster recovery is worth closing regardless, and `vault-reinitialize` is the case where recovery is least comfortable. But please say what is actually lost — a re-run, not the key — because a comment that overstates the stakes is the thing someone later checks and stops trusting.

Everything else stands: the three-workflow sweep is the sibling-drift class caught properly, the sync-to-sops reasoning about legitimate `skipped` is correct and well argued, and the CANNOT-DETERMINE arm is there.

Not blocking on the wording. Fix the gate, add the arm, and I will merge.

jonhill90:
Verified against the branch, not the PR body. Both findings are closed and the rework is better than what I asked for.

Ran the control locally on `4763292d`:

```
1..8
ok 1 CONTROL: the check PASSES on the real, unmutated workflows
ok 2 removing the copy-back step's id turns it red
ok 3 reverting the revert step's if: to a bare always() turns it red
ok 4 THE ASSERTION THAT MATTERS: a revert gated on != 'failure' (admits a SKIPPED copy-back) turns it red
ok 5 removing the revert step's if: condition entirely turns it red
ok 6 a revert step with no preceding copy-back step at all turns it red
ok 7 the check refuses to pass vacuously when it can see no workflows
ok 8 THE ASSERTION THAT MATTERS: every revert step losing its recognisable shape reports CANNOT DETERMINE, not PASS
```

Arm 4 is the one I asked for and it is genuinely red on the exact window — the previous revision would have passed it.

The gates are right: `== 'success'` on init and reinitialize where a skip is never legitimate, and `(steps.compare.outputs.changed != 'true' || steps.copy-back.outcome == 'success')` on sync-to-sops, which preserves the no-drift skip you correctly argued for without also admitting an upstream death.

The reworded comment is more useful than the correction I asked for. I said "say what is actually lost — a re-run, not the key"; you went further and said *which* re-run and why the checkout is the thing worth keeping: "What reverting early would discard is this attempt's mutated checkout — the thing worth inspecting or retrying from — not the key's only copy." That is the version someone can act on at 3am.

Merging.

PR #781, opened 2026-08-05
fix(deploy): serialise the deploy jobs — two at once broke twice in an hour

jonhill90:
Closes #778.

## Plan

Every deploy job SSHes to the same VPS, works in the same checkout (`/opt/hill90/app`) and deploys onto the same running estate. `deploy.yml` already had a `deploy-db → deploy-auth → deploy-minio` chain. `deploy-vault` and `deploy-observability` hung off `changes` directly, so they ran *alongside* it.

Both of 2026-08-05's deploy failures follow from exactly that:

| Run | What ran together | Failure |
|---|---|---|
| `31039334078` | `deploy-db` and `deploy-observability`, both started `19:25:45Z` | `error: fetching ref refs/remotes/origin/main failed: incorrect old value provided` |
| `31040415190` | `deploy-auth` and `deploy-observability` | `no Keycloak login form in the response` |

The first left a **partial deploy**: `deploy-auth` was skipped after db died, so of h#774's three changed compose files only `observability.yml` reached production, and the whole thing surfaced as one failed job. The second **named the wrong service** — Grafana was healthy the whole time (`ok Grafana is serving`, `ok Grafana redirected to https://auth.hill90.com, realm platform`), and the login form serves correctly now; auth's Keycloak restart had simply landed underneath the assertion.

**The two contend on different resources** — the git checkout, and the running estate. That distinction rules out one of the three candidate fixes in #778: per-job checkout directories fix the first and leave the second untouched. Serialising fixes both, and does it by extending a chain that already existed rather than by introducing a locking mechanism.

`deploy-observability` goes **last** on purpose: its post-deploy checks assert against the other services, so it must not start until they have settled.

## Risks

Two, and both are real rather than pro-forma.

- **A five-service deploy is now serial.** It takes roughly the sum of its parts instead of the slowest branch. Accepted: a homelab deploy that is a few minutes slower and correct beats a fast one that half-applies.
- **A db failure now skips vault and observability**, which previously deployed anyway — the chain carries `!failure()`. I read that as the right behaviour (a failed database deploy should not be followed by four more), but it **is** a behaviour change and it is yours to veto. If you would rather keep them independent of db's outcome, the alternative is a VPS-side `flock`, which serialises without coupling outcomes — at the cost of jobs that sit burning runner minutes waiting on the lock, and no protection for anything that deploys outside this workflow.

Not addressed here: nothing serialises this workflow against a *human* running `deploy.sh` on the VPS directly. `concurrency: deploy-prod` only covers runs of this workflow.

## Rollback

Revert the commit. `deploy-vault` and `deploy-observability` return to hanging off `changes`, and both failure modes come back.

## Validation Evidence

Positive control, with both real failures replayed as separate mutations:

```
$ bats tests/scripts/deploy-jobs-serialized-control.bats
1..8
ok 1 CONTROL: passes on the real tree
ok 2 CONTROL: it counts the real jobs, not zero
ok 3 observability parallel to the db chain — the git-race run — is caught
ok 4 vault parallel to the chain — the Keycloak-restart run — is caught
ok 5 a fork — two jobs sharing an ancestor but not each other — is caught
ok 6 a different but still-serial order passes
ok 7 a workflow with no deploy-* jobs is CANNOT DETERMINE, not a pass
ok 8 ci.yml genuinely invokes this check, not just mentions it
```

Test 5 is the one an eyeball misses: two jobs that each have a `needs` and *look* chained, while neither is reachable from the other. The check asserts the property — for every pair of deploy jobs, one must be an ancestor of the other — not a literal graph, which is why test 6 requires a reordering that stays serial to **pass**. This check should not have to be edited every time the dependency story changes.

Resulting graph:

```
changes
deploy-db              needs=['changes']
deploy-auth            needs=['changes', 'deploy-db']
deploy-minio           needs=['changes', 'deploy-auth']
deploy-vault           needs=['changes', 'deploy-minio']
deploy-observability   needs=['changes', 'deploy-vault']
```

Both deploy checks pass on the real tree, and #779's check still passes alongside this one:

```
$ python3 scripts/checks/check_deploy_jobs_serialized.py
PASS — the 5 deploy jobs form a single chain; no two
       can be in flight against the VPS at once.
$ python3 scripts/checks/check_dispatch_targets_one_service.py
PASS — change detection is gated to non-dispatch events, so a manual
       deploy is decided by the `service` input alone.
```

**What this evidence does not cover.** The check is static. It proves no two deploy jobs *can* be scheduled together; it does not prove the serialised deploy still works end to end. None of the four files here is on `deploy.yml`'s `on.push.paths`, so merging deploys nothing — the runtime proof has to be the next real multi-service deploy, and I will post its job timings here when one happens rather than claim it now.

## Estate state

Verified before opening this, after the last of today's deploys:

```
missing: none  unhealthy:0  restarting:0  tenant:7
  hill90.com 200
  auth.hill90.com/realms/platform 200
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,12 @@ jobs:
       - name: Assert a manual deploy targets one service
         run: python3 scripts/checks/check_dispatch_targets_one_service.py
 
+      # #778: deploy-db and deploy-observability started at the same second and
+      # raced on /opt/hill90/app; an hour later deploy-auth restarted Keycloak
+      # underneath observability's login check. Hermetic — reads deploy.yml.
+      - name: Assert no two deploy jobs can run at once
+        run: python3 scripts/checks/check_deploy_jobs_serialized.py
+
       - name: Install bats
         run: |
           sudo apt-get update
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -104,17 +104,31 @@ jobs:
       service: minio
     secrets: inherit
 
+  # SERIALISED, and this is load-bearing rather than tidiness. Every deploy
+  # job SSHes to the same VPS, works in the same checkout (/opt/hill90/app)
+  # and deploys onto the same running estate. On 2026-08-05 two jobs in
+  # parallel broke twice in an hour: run 31039334078 raced on the git
+  # checkout ("incorrect old value provided") and left a PARTIAL deploy
+  # reported as a single failed job; run 31040415190 restarted Keycloak
+  # underneath the observability job's Grafana login check, which then
+  # blamed the wrong service. See #778.
+  #
+  # deploy-observability is deliberately LAST: its post-deploy checks assert
+  # against the other services, so it must not start until they have settled.
+  #
+  # The cost is real and accepted — a five-service deploy is now serial, not
+  # parallel. scripts/checks/check_deploy_jobs_serialized.py holds the line.
   deploy-vault:
-    needs: [changes]
-    if: needs.changes.outputs.vault == 'true' || github.event.inputs.service == 'all' || github.event.inputs.service == 'vault'
+    needs: [changes, deploy-minio]
+    if: always() && (needs.changes.outputs.vault == 'true' || github.event.inputs.service == 'all' || github.event.inputs.service == 'vault') && !failure() && !cancelled()
     uses: ./.github/workflows/reusable-deploy-service.yml
     with:
       service: vault
     secrets: inherit
 
   deploy-observability:
-    needs: [changes]
-    if: needs.changes.outputs.observability == 'true' || github.event.inputs.service == 'all' || github.event.inputs.service == 'observability'
+    needs: [changes, deploy-vault]
+    if: always() && (needs.changes.outputs.observability == 'true' || github.event.inputs.service == 'all' || github.event.inputs.service == 'observability') && !failure() && !cancelled()
     uses: ./.github/workflows/reusable-deploy-service.yml
     with:
       service: observability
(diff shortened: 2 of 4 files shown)


jonhill90:
**Runtime proof, closing the gap this PR body flagged as unproven.**

The evidence section said the check is static — it proves no two deploy jobs *can* be scheduled together, not that a serialised deploy still works end to end — and that the proof had to be a real multi-service deploy. Ran one: `gh workflow run deploy.yml -f service=all --ref main`, run `31043850839`, **completed success**.

```
changes                 success  20:23:46 -> 20:23:52
deploy-db               success  20:24:04 -> 20:24:39
deploy-auth             success  20:24:50 -> 20:26:19
deploy-minio            success  20:26:24 -> 20:26:55
deploy-vault            success  20:26:58 -> 20:27:24
deploy-observability    success  20:27:29 -> 20:28:56
```

Every job starts *after* the previous one finished — **zero overlap**, which is the property the check asserts statically and this demonstrates in the running system. Compare run `31039334078`, where `deploy-db` and `deploy-observability` both started at `19:25:45Z`.

Two things worth naming beyond "it passed":

**observability's Grafana→Keycloak login check passed.** That is the check that failed in run `31040415190` when auth restarted Keycloak underneath it. Here auth finished at 20:26:19 and observability did not start until 20:27:29 — 70 seconds later, and after minio and vault in between. The ordering did the work it was put there to do.

**The cost is now measured rather than estimated.** Five services, strictly serial, **5m10s** end to end (20:23:46 → 20:28:56). The PR said "roughly the sum of its parts" and accepted it; that is what the sum is.

Estate verified immediately after, unchanged:

```
missing: none  unhealthy:0  restarting:0  tenant:7
  hill90.com 200
  auth.hill90.com/realms/platform 200
```

This was a real production deploy of all five platform services, through the pipeline, re-applying config already at `main`'s HEAD.

PR #783, opened 2026-08-05
fix(minio): secret_for() couldn't distinguish a decrypt failure from an absent key

jonhill90:
## scripts/ silent-success sweep, finding 3

\`secret_for()\`'s \`sops\` call piped stderr to \`/dev/null\` and any decrypt failure (bad/missing age key, corrupted store) resolved to the same empty string as a genuinely-absent key — both produced the identical \`"not set and was not found in \${SECRETS_FILE}"\` message, sending an operator debugging the wrong file instead of their age key. \`die()\` still fired either way; this was never a silent-success bug, only a misleading one.

**The fix.** Split the decrypt from the extraction:
\`\`\`bash
local decrypted
if ! decrypted=$(sops -d "$SECRETS_FILE" 2>/dev/null); then
    die "could not decrypt ${SECRETS_FILE} — check the age key for prod"
fi
\`\`\`
then grep the key out of \`$decrypted\` separately. Deliberately **not** \`local decrypted=$(sops -d ...)\` — that natural-looking form discards sops's own exit status, since \`local\` returns its own status and \`set -e\` never sees the command substitution's failure inside it. That's exactly shape 1 of this same sweep (a pipeline/assignment whose exit status comes from the wrong command), so writing the fix that way would reintroduce the defect while fixing it.

**Two constraints honored.** No secret value is ever printed — \`$decrypted\` is only ever piped into \`grep\`, never echoed or included in any message, including the new failure path. The new message names the file and the failure mode (\`"could not decrypt \${SECRETS_FILE} — check the age key for prod"\`), never the value.

**Positive-controlled, both directions**, with a real forced \`sops\` failure — a harness sourcing the real \`_common.sh\` and the real \`secret_for()\` (extracted from the file with \`sed\`, not reimplemented) with a stubbed \`sops\`. Old code: decrypt-fails and key-absent cases produce the byte-identical message, confirmed by diffing the two outputs directly. New code: distinct messages; confirmed the fixed source still resolves a real key correctly; confirmed nothing a stubbed \`sops\` writes to its own stderr ever reaches the caller's output.

**New \`tests/scripts/minio-secret-for-control.bats\`**, 5 tests, all reverted against the pre-fix source first to confirm the 2 relevant ones fail for exactly the predicted reason, then restored and reconfirmed green.

**Full suites**: \`bats tests/scripts/\` 623 passed, 0 failed. \`pytest tests/checks/\` 81 passed, 1 failed (\`test_tenant_privilege_boundary_holds\`) — confirmed unrelated (a container-resource-contention flake) and passes 2/2 clean rerun alone.

## Test plan
- [x] Positive control: real forced \`sops\` failure, old code shown producing the identical message as key-absent
- [x] New code confirmed to distinguish the two cases
- [x] No secret value ever printed, in any path — verified with a stubbed \`sops\` that would leak if the redirect broke
- [x] Full \`bats tests/scripts/\` — 623 passed, 0 failed
- [x] \`pytest tests/checks/\` — 81 passed, 1 unrelated flake confirmed clean alone

### DIFF
--- a/scripts/minio.sh
+++ b/scripts/minio.sh
@@ -60,7 +60,21 @@ secret_for() {
     if [ -z "$value" ]; then
         require_file "$SECRETS_FILE" "Secrets file"
         ensure_age_key prod
-        value=$(sops -d "$SECRETS_FILE" 2>/dev/null | grep "^${var}=" | cut -d= -f2- || true)
+        # Split declaration from assignment on purpose: `local decrypted=$(...)`
+        # would discard sops's own exit status — `local` returns its own status,
+        # and set -e does not see the command substitution's failure inside it.
+        # That is exactly the shape this fix exists to close, so writing it the
+        # natural way would reintroduce the defect while fixing it.
+        local decrypted
+        if ! decrypted=$(sops -d "$SECRETS_FILE" 2>/dev/null); then
+            die "could not decrypt ${SECRETS_FILE} — check the age key for prod"
+        fi
+        # A decrypt failure and a genuinely-absent key used to produce the
+        # identical "not set and was not found" message, sending an operator
+        # debugging the wrong file instead of their age key. Never echo
+        # $decrypted itself — only pipe it, so the file's contents never reach
+        # stdout/stderr regardless of which branch below fires.
+        value=$(printf '%s' "$decrypted" | grep "^${var}=" | cut -d= -f2- || true)
     fi
     [ -n "$value" ] || die "${var} is not set and was not found in ${SECRETS_FILE}"
     printf '%s' "$value"
--- a/tests/scripts/minio-secret-for-control.bats
+++ b/tests/scripts/minio-secret-for-control.bats
@@ -1,0 +1,96 @@
+#!/usr/bin/env bats
+
+# POSITIVE CONTROL for scripts/minio.sh's secret_for().
+#
+# A sops decrypt failure (bad/missing age key, corrupted store) and a
+# genuinely-absent key used to print the IDENTICAL "not set and was not
+# found" message, sending an operator debugging the wrong file instead of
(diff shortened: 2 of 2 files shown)


jonhill90:
Verified against the branch and against `origin/main`, and the control does what a control has to do. Run locally on `99b40ba0`, then again with `scripts/minio.sh` swapped back to `origin/main`:

```
                                                              fixed   pre-fix
1 CONTROL: resolves a real key from a successful decrypt        ok      ok
2 THE ASSERTION THAT MATTERS: a decrypt failure names the file  ok      not ok
3 CONTROL: a genuinely-absent key still gets not-found          ok      ok
4 CONTROL: the two failure messages are distinct                ok      not ok
5 no decrypted content is ever printed on a decrypt failure     ok      ok
```

Exactly two arms flip, and they are the two that encode the defect. The three that stay green are the ones proving the fix did not change behaviour it should not have touched — a control where everything flips would not distinguish a fix from a rewrite.

**You avoided the trap.** `local decrypted` is declared on its own line and the comment says why, in the same terms as the defect itself: `local decrypted=$(...)` returns `local`'s status, not `sops`'s, so writing it the natural way would have reintroduced the exact shape the fix closes. That is the one thing I was watching for.

**And the secret never moves.** `$decrypted` is only ever piped, never echoed, on either branch — and arm 5 asserts it rather than leaving it to review.

One small thing, not blocking and not worth a push: the `sed -n '/^secret_for() {/,/^}/p'` extraction has no guard that it found anything. If the function is ever renamed, `eval ""` succeeds and `secret_for` is simply undefined. That fails the tests rather than passing them vacuously — status 127 and no matching output — so the 0==0 trap does not bite here, but a one-line assertion that the extraction is non-empty would make the failure say *why*. Worth adding next time this file is open.

Merging.

PR #784, opened 2026-08-05
fix(vault): status checks could not distinguish sealed from unreachable

jonhill90:
## Summary
- `cmd_init`, `cmd_unseal` and `cmd_bootstrap_approles` all inferred OpenBao's state via `bao status -format=json | grep '"x"' | tr -d ' ,"' || echo ""`, with no `pipefail` — a genuinely-false field and an unreachable OpenBao both produced an empty string, so each site silently proceeded into real init/unseal logic instead of reporting the actual problem.
- `bao status` exits 0 unsealed, 2 sealed (both print full JSON) or 1 on error (no JSON) — verified empirically against a running container. Each call site now captures `bao_exec`'s own exit code directly and dies on anything but 0/2, naming reachability instead of leaving the caller to hit an unrelated downstream error later.

## Test plan
- [x] `shellcheck --severity=error scripts/vault.sh` clean
- [x] `bash -n scripts/vault.sh` clean
- [x] New positive control `tests/scripts/vault-status-exit-code-control.bats` (7 tests) extracts the real functions from the real file and stubs only `bao_exec` — 3 "assertion that matters" tests, 4 control tests
- [x] Reverted the fix and reran: the 3 assertion tests go red for the predicted reason (old code plows into real logic instead of dying cleanly), all 4 control tests stay green — confirms the defect reproduces and the fix is what closes it
- [x] Restored the fix, reran: all 7 green
- [x] Full `bats tests/scripts/*.bats`: 625/625 passed
- [x] Full `pytest tests/checks/`: 82/82 passed

### DIFF
--- a/scripts/vault.sh
+++ b/scripts/vault.sh
@@ -113,9 +113,22 @@ cmd_init() {
     echo "================================"
     echo ""
 
-    # Check if already initialized
+    # Check if already initialized. `bao status` exits 0 unsealed, 2 sealed —
+    # both mean it actually answered and the JSON is usable — or 1 on error
+    # (unreachable, container broken). The old form piped straight into
+    # grep/tr with no pipefail, so a genuine "sealed: false" and an
+    # unreachable OpenBao both collapsed to the same empty string, and the
+    # code proceeded down the "not initialized" branch either way. Use the
+    # exit code, which distinguishes them, instead of inferring from output
+    # that looks the same in both cases.
+    local status_json rc=0
+    status_json=$(bao_exec status -format=json 2>&1) || rc=$?
+    case "$rc" in
+        0|2) ;;
+        *) die "could not determine OpenBao's status (bao status exited ${rc}) — is the container reachable? ${status_json}" ;;
+    esac
     local init_status
-    init_status=$(bao_exec status -format=json 2>/dev/null | grep '"initialized"' | tr -d ' ,"' || echo "")
+    init_status=$(printf '%s' "$status_json" | grep '"initialized"' | tr -d ' ,"')
     if [[ "$init_status" == *"true"* ]]; then
         warn "OpenBao is already initialized"
         echo "Use 'vault.sh unseal' if it is sealed, or 'vault.sh status' to check."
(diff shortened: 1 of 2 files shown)


jonhill90:
Verified against the branch and `origin/main`. Ran the control both ways:

```
                                                                fixed   pre-fix
1 cmd_init: unreachable (rc 1) dies naming reachability           ok     not ok
2 cmd_init: CONTROL already initialized (rc 0)                    ok     ok
3 cmd_unseal: unreachable (rc 1) dies naming reachability         ok     not ok
4 cmd_unseal: CONTROL already unsealed (rc 0)                     ok     ok
5 bootstrap_approles: unreachable dies before calling cmd_unseal  ok     not ok
6 bootstrap_approles: CONTROL sealed (rc 2) calls cmd_unseal      ok     ok
7 bootstrap_approles: CONTROL unsealed (rc 0) skips cmd_unseal    ok     ok
```

Exactly the three defect assertions flip and all four controls hold on both sides. Arm 6 is the one I would have most wanted to see: rc 2 must still mean "sealed, carry on and unseal", because a fix that treated any non-zero as unreachable would break the normal boot path — and that is the shape this change could plausibly have got wrong.

The exit-code approach is right. `bao status` answering 0 or 2 both mean it *answered* and the JSON is usable; only 1 means nothing came back. Inferring from an empty grep result could never separate those.

**One thing I got wrong while reviewing, worth recording because it is the same trap this PR is about.** I flagged that dropping `|| echo ""` would let a failed `grep` abort under `set -e` (line 5). Tested it instead of trusting the reasoning:

```
set -e
v=$(printf '%s' '{"sealed":false}' | grep '"initialized"' | tr -d ' ,"')
echo "REACHED, v=[$v]"   ->  REACHED, v=[]   exit=0
```

It does not abort, because the pipeline ends in `tr`, which succeeds on empty input, so `grep`'s failure is masked. Which is *exactly* category 1 of your own sweep — a pipeline whose exit status comes from the last command rather than the one doing the work. The removal is safe, and the reason it is safe is the defect shape you went looking for.

**One note, not blocking.** `status_json=$(bao_exec status -format=json 2>&1)` folds stderr in, and `${status_json}` is interpolated into the `die` message. `bao status` output carries no credential material, so this is fine today — but it is a place where a future error string could surface something an operator pastes into a ticket. If it ever grows, truncate or drop the interpolation rather than the `2>&1`, which is genuinely useful for diagnosing reachability.

Merging.

PR #785, opened 2026-08-05
fix(backup): cmd_prune() counted a backup as pruned even when rm failed

jonhill90:
## Summary
- `cmd_prune()`'s `rm -rf "$backup_dir"` exit status was unchecked and `pruned` incremented unconditionally, so a removal that failed (permissions, a read-only mount, a file locked elsewhere) was still reported as pruned — an operator reading "Pruned N backup(s)" had no way to know the backup was still on disk.
- Named family per the `scripts/` silent-success sweep, kept as a standalone third PR rather than parked, per instruction.

## Test plan
- [x] `shellcheck --severity=error scripts/backup.sh` clean
- [x] `bash -n scripts/backup.sh` clean
- [x] New positive control `tests/scripts/backup-prune-exit-status-control.bats` (3 tests) extracts the real `cmd_prune` and forces a real `rm -rf` failure via a read-only parent directory — not a stub
- [x] Reverted the fix and reran: the "assertion that matters" test goes red for the predicted reason, both control tests stay green — confirms the defect reproduces and the fix is what closes it
- [x] Restored the fix, reran: all 3 green
- [x] Full `bats tests/scripts/*.bats`: 633/633 passed
- [x] Full `pytest tests/checks/`: 82/82 passed

### DIFF
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -748,8 +748,11 @@ cmd_prune() {
 
             if [ "$backup_epoch" -lt "$cutoff_epoch" ]; then
                 echo "  Removing: ${svc}/${ts}"
-                rm -rf "$backup_dir"
-                pruned=$((pruned + 1))
+                if rm -rf "$backup_dir"; then
+                    pruned=$((pruned + 1))
+                else
+                    warn "failed to remove ${svc}/${ts} — not counting it as pruned"
+                fi
             fi
         done
     done
--- a/tests/scripts/backup-prune-exit-status-control.bats
+++ b/tests/scripts/backup-prune-exit-status-control.bats
@@ -1,0 +1,74 @@
+#!/usr/bin/env bats
+
+# POSITIVE CONTROL for scripts/backup.sh's cmd_prune().
+#
+# The old code ran `rm -rf "$backup_dir"` and unconditionally incremented
+# `pruned`, so a removal that failed (permissions, a mount gone read-only,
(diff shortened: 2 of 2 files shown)


jonhill90:
Verified against `origin/main`. Ran the control both ways:

```
                                                              fixed   pre-fix
1 THE ASSERTION THAT MATTERS: a failed removal is not counted   ok     not ok
2 CONTROL: a successful removal is counted and the dir is gone  ok     ok
3 CONTROL: a backup inside the retention window is left alone   ok     ok
```

Exactly one arm flips and it is the one encoding the defect. Arm 3 is the one I would have gone looking for if it were missing: a fix in a prune loop can very easily start deleting things it should not, and a control that only proves "failures are not counted" would not have noticed.

**The technique is the best part.** The control forces a *real* `rm -rf` failure — a read-only parent directory, so the entry genuinely cannot be unlinked — rather than mocking `rm`. A mock would have proved the `if` branches correctly on a fake exit code; this proves it branches correctly on the thing that actually happens. And it extracts the real `cmd_prune` from the real file, so it is not testing a reimplementation.

The change also does the right thing on failure rather than merely not-counting: `warn` names which backup survived, so an operator sees "Pruned 0" *and* the reason, instead of a silently smaller number.

I raised this one over your "hold it" judgement earlier, and I still think that was right — but your severity read was accurate and worth preserving: nothing downstream treats `pruned` as a safety signal, so this was never a data-loss risk. It was an operator being told a number that was not true.

Merging.

